PR #9, opened 2019-03-14
Uses `git checkout` instead of `git reset` for `spack_commit` checkout

### LINKED ISSUE #8: Checking out a custom spack branch
When ``spack_commit`` is present in the ``project.json`` file, uberenv uses that commit of spack.

It currently uses ``git reset --hard $commit``, which resets the branch pointer for spack's develop branch to the desired commit.  Unfortunately, this makes it look like it is still using the develop branch, which could be confusing to users.

Instead, it should use ``git checkout $commit``, which puts spack in a detached head state indicating the current commit.

kennyweiss:
Resolves #8.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -428,7 +428,7 @@ def main():
             sha1 = project_opts["spack_commit"]
             print "[info: using spack commit %s]" % sha1
             os.chdir(pjoin(dest_dir,"spack"))
-            sexe("git reset --hard %s" % sha1,echo=True)
+            sexe("git checkout %s" % sha1,echo=True)
 
     if opts["spack_pull"]:
         # do a pull to make sure we have the latest


cyrush:
checkout can fail in cases where reset will not due to dirty state, i think thats why i have it there

kennyweiss:
Thanks @cyrush.  In this case though, we just cloned a clean spack.  
Are there possible failure cases here?

With this change, it seems much easier (to me) to understand which commit our local spack is based on.

cyrush:
I see this only happens right after clone. I definitely had a reason but the logic may have evolved to note be needed.

cyrush:
Had some time to ponder, i think this is a good improvement. 
I wanted to check the behavior when combined with the uberenv --pull argument. 

pull should not change anything when we use a specific commit. 

I did a quick check using git directly,  from my test I believe we are good

cyrush:
thanks @kennyweiss!

PR #19, opened 2020-01-30
Minor tweaks

white238:
* Allow package_final_phase to be undefined (REMOVED)
* Catch errors in spack activate

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -70,7 +70,7 @@ from os.path import join as pjoin
 def sexe(cmd,ret_output=False,echo = False):
     """ Helper for executing shell commands. """
     if echo:
-        print("[exe: {}]".format(cmd))
+        print("[exe: {0}]".format(cmd))
     if ret_output:
         p = subprocess.Popen(cmd,
                              shell=True,
@@ -205,7 +205,7 @@ def parse_args():
     if not opts["spack_config_dir"] is None:
         opts["spack_config_dir"] = os.path.abspath(opts["spack_config_dir"])
         if not os.path.isdir(opts["spack_config_dir"]):
-            print("[ERROR: invalid spack config dir: {} ]".format(opts["spack_config_dir"]))
+            print("[ERROR: invalid spack config dir: {0} ]".format(opts["spack_config_dir"]))
             sys.exit(-1)
     return opts, extras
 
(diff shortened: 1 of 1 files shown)


uberenv.py
```diff
@@ -465,7 +468,7 @@ def install(self):
         if self.opts["ignore_ssl_errors"]:
             install_cmd += "-k "
         install_cmd += "dev-build -d {} ".format(self.pkg_src_dir)
-        if not self.opts["install"]:
+        if not self.opts["install"] and self.pkg_final_phase:
```
adrienbernede: At this point, shouldn’t we inform the user that "not install" mode expects a final phase? Or document the behavior?

Uberenv will continue and actually build the project in spack using the current source. The behavior is thus slightly different from what is used to be: build a fake project.

white238: I am not sure what the correct answer is here. If you don't have a final phase and call uberenv (without --install) do we want it to fail?

Right now everyone I know has a final phase but I am not sure that's going to be the case forever.  I was just working around splitting the devtools for my two projects out and this seemed the best thing to do since the devtools bundlepackage doesn't have any phases.

Thoughts @cyrush ?

cyrush: for now I would opt for an error if --install is not selected and there is no final phase.

It's not clear what we are asking uberenv to do in that case.

white238: I am fine with whatever.  I hide all this in scripts.  I was just wondering what we feel is the best workflow for users.

To be clear, if we have a project/package foo that does not have a final phase.  We want it to error out unless you say --install.  Instead of defaulting to just building everything?

Is this what you want?

uberenv.py foo
 Error: No final phase declared for package 'foo'.  Either declare what phase to stop after on the command line or in the project.json.  Alternately, the command line option "--install" will build your complete package.

adrienbernede: My 2 cents...

@white238 I understand you concern. It does not surprise me much however since the default behavior of uberenv _is not_ --install. To me, this choice means that Uberenv is primarily meant to deal with dependency installs.

I’ll let the creator answer this question though.

white238:
Frankly, I forgot about this PR.

I removed the change in the --install logic due to concerns.  I also changed all the .format's to be explicit in their indexing due to me having run into weird compatibility problems in the past...

adrienbernede:
I quite forgot about it too... I was focusing of mine, on which I would use some feedback since I think it’s ready (except for some format not indexed ;) ).

white238:
Closed and moved to #22 to avoid large set of conflicts.

PR #21, opened 2020-03-10
Fixes, Improvements and CI

adrienbernede:
Fixes some undesired behavior introduced lately (see https://github.com/LLNL/conduit/pull/502).

Also introduce some robustness improvements:
 - Check if a spec is already installed.
 - Ability to use the spec hash to accurately retrieve information (e.g. spack install path).

Add CI triggering Gitlab CI pipeline in Conduit and Serac.

Recent change:
 - Roll back of sexe() implementation after feedback.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -52,3 +52,11 @@ trigger-conduit:
     project: radiuss/conduit
     branch: feature/uberenv_ci
     strategy: depend
+
+trigger-serac:
+  variables:
+    UPDATE_UBERENV: $CI_COMMIT_REF_NAME
+  trigger:
+    project: bernede1/serac
+    branch: feature/up-uber
+    strategy: depend
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,54 @@
+###############################################################################
+# Copyright (c) 2014-2020, Lawrence Livermore National Security, LLC.
+#
+# Produced at the Lawrence Livermore National Laboratory
+#
+# LLNL-CODE-666778
+#
+# All rights reserved.
+#
+# This file is part of Conduit.
+#
+# For details, see https://lc.llnl.gov/conduit/.
+#
+# Please also read conduit/LICENSE
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the disclaimer below.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the disclaimer (as noted below) in the
+#   documentation and/or other materials provided with the distribution.
+#
+# * Neither the name of the LLNS/LLNL nor the names of its contributors may
+#   be used to endorse or promote products derived from this software without
+#   specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL LAWRENCE LIVERMORE NATIONAL SECURITY,
+# LLC, THE U.S. DEPARTMENT OF ENERGY OR CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+###############################################################################
+
+# Triggers Conduit pipeline asking for uberenv update
+# UPDATE_UBERENV variable, when set, will tell conduit to use the value as
+# a reference to the uberenv env revision to checkout on github.
+trigger-conduit:
+  variables:
+    UPDATE_UBERENV: $CI_COMMIT_REF_NAME
+  trigger:
+    project: radiuss/conduit
+    branch: feature/uberenv_ci
+    strategy: depend
--- a/uberenv.py
+++ b/uberenv.py
@@ -282,6 +282,7 @@ class SpackEnv(UberEnv):
         self.pkg_src_dir = self.set_from_args_or_json("package_source_dir")
 
         self.spec_hash = ""
+        self.use_install = False
 
         # Some additional setup for macos
         if is_darwin():
@@ -333,7 +334,7 @@ class SpackEnv(UberEnv):
             sys.exit(-1)
 
 
-    def find_spack_pkg_path_form_hash(self, pkg_name, pkg_hash):
+    def find_spack_pkg_path_from_hash(self, pkg_name, pkg_hash):
         r,rout = sexe("spack/bin/spack find -p /{}".format(pkg_hash), ret_output = True)
         for l in rout.split("\n"):
             if l.startswith(pkg_name):
@@ -445,8 +446,8 @@ class SpackEnv(UberEnv):
             # let spack try to auto find compilers
             sexe("spack/bin/spack compiler find", echo=True)
 
+        # hot-copy our packages into spack
         if self.pkgs:
-            # hot-copy our packages into spack
             dest_spack_pkgs = pjoin(spack_dir,"var","spack","repos","builtin","packages")
             sexe("cp -Rf {} {}".format(self.pkgs,dest_spack_pkgs))
 
@@ -469,21 +470,27 @@ class SpackEnv(UberEnv):
                         res = sexe(unist_cmd, echo=True)
 
     def show_info(self):
-        spec_cmd = "spack/bin/spack spec -IL " + self.pkg_name + self.opts["spec"]
+        # prints install status and 32 characters hash
+        options="--install-status --very-long"
+        spec_cmd = "spack/bin/spack spec {0} {1}{2}".format(options,self.pkg_name,self.opts["spec"])
 
         res, out = sexe(spec_cmd, ret_output=True, echo=True)
         print(out)
 
+        #Check if spec is already installed
         for line in out.split("\n"):
+            # Example of matching line: ("status"  "hash"  "package"...)
+            # [+]  hf3cubkgl74ryc3qwen73kl4yfh2ijgd  serac@develop%clang@10.0.0-apple~debug~devtools~glvis arch=darwin-mojave-x86_64
             if re.match(r"^(\[\+\]| - )  [a-z0-9]{32}  " + re.escape(self.pkg_name), line):
                 self.spec_hash = line.split("  ")[1]
+                # if spec already installed
                 if line.startswith("[+]"):
-                    pkg_path = self.find_spack_pkg_path_form_hash(self.pkg_name,self.spec_hash)
+                    pkg_path = self.find_spack_pkg_path_from_hash(self.pkg_name,self.spec_hash)
                     install_path = pkg_path["path"]
                     if os.path.isdir(install_path):
-                        print("[Warning: {} {} has already been install in {}]".format(self.pkg_name, self.opts["spec"],install_path))
+                        print("[Warning: {} {} has already been installed in {}]".format(self.pkg_name, self.opts["spec"],install_path))
                         print("[Warning: Uberenv will proceed using this directory]".format(self.pkg_name))
-                        self.opts["use_install"] = True
+                        self.use_install = True
 
         return res
 
@@ -491,7 +498,7 @@ class SpackEnv(UberEnv):
         # use the uberenv package to trigger the right builds
         # and build an host-config.cmake file
 
-        if not "use_install" in self.opts:
+        if not self.use_install:
             install_cmd = "spack/bin/spack "
             if self.opts["ignore_ssl_errors"]:
                 install_cmd += "-k "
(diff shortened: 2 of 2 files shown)


adrienbernede:
LGTM

adrienbernede:
LGTM

adrienbernede:
LGTM

adrienbernede:
LGTM

adrienbernede:
LGTM

adrienbernede:
LGTM

uberenv.py
```diff
@@ -493,32 +521,51 @@ def install(self):
         # note: this assumes package extends python when +python
         # this may fail general cases
         if self.opts["install"] and "+python" in full_spec:
-            activate_cmd = "spack/bin/spack activate " + self.pkg_name
+            activate_cmd = "spack/bin/spack activate " + self.pkg_name + self.opts["spec"]
```
kennyweiss: Since ``spack`` commands now need additional context/environments, would it make sense to go through a utility function, e.g. ``create_spack_command()`` that deals with the path to ``spack`` and adding the appropriate environment? This could help improve consistency.

This could definitely be handled in a separate PR.

uberenv.py
```diff
@@ -493,32 +521,51 @@ def install(self):
...
             else:
-                pkg_lnk_dir = "{}-install".format(self.pkg_name)
-                if os.path.islink(pkg_lnk_dir):
-                    os.unlink(pkg_lnk_dir)
-                print("")
-                print("[symlinking install to {}]").format(pjoin(self.dest_dir,pkg_lnk_dir))
-                os.symlink(pkg_path["path"],os.path.abspath(pkg_lnk_dir))
-                hcfg_glob = glob.glob(pjoin(pkg_lnk_dir,"*.cmake"))
+                # Symlink host-config file
+                hcfg_glob = glob.glob(pjoin(pkg_path["path"],"*.cmake"))
```
kennyweiss: Suggestion: Perhaps ``hc`` would be clearer than ``hcfg``  for this prefix? (I had trouble figuring it out without the comment on line 534)

uberenv.py
kennyweiss: This comment might be more appropriate in the context of https://github.com/LLNL/uberenv/pull/18

uberenv.py
```diff
@@ -67,18 +67,30 @@
 from os.path import join as pjoin
 
 
-def sexe(cmd,ret_output=False,echo = False):
+def sexe(cmd, ret_output=False, print_output=True, echo=False):
```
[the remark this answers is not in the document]
cyrush: I have maybe a hundred copies of this function :-)

for this use case: 
if folks want to print the output, I would recommend using ret_output=True, and then printing the result outside of the this function. 

Maybe I am just very superstitious, but not using  `res = p.communicate()[0]  `  has deadlocked for me in the distant past.

.gitlab-ci.yml
```diff
@@ -0,0 +1,54 @@
...
+###############################################################################
+
+# Triggers Conduit pipeline asking for uberenv update
+# UPDATE_UBERENV variable, when set, will tell conduit to use the value as
+# a reference to the uberenv env revision to checkout on github.
+trigger-conduit:
+  variables:
+    UPDATE_UBERENV: $CI_COMMIT_REF_NAME
+  trigger:
+    project: radiuss/conduit
```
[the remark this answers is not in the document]
adrienbernede: I’m having a hard time validating Uberenv for every projects using it, and there are not so many.
This is both because there is no test in Uberenv per se, and because updating and testing it in other project is not automated.

What this does is to trigger conduit CI pipeline in Gitlab, asking for uberenv to be updated.
How it works:
In conduit, I added a script in (GitLab) CI that will be used only if `UPDATE_UBERENV` is set. The value is supposed to be the branch name in Uberenv repo.
Now if I trigger Conduit pipeline with this variable, Conduit install will be tested with the specified version of uberenv. I intend to do the same for other projects using Uberenv, but the mechanism can be use to test newer version of spack, or cross-project integration in general.

Note: `branch` will be set to master once feature/uberenv_ci will be merge in conduit. This PR is thus connected to https://github.com/LLNL/conduit/pull/517

.gitlab-ci.yml
```diff
@@ -0,0 +1,54 @@
+###############################################################################
+# Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
```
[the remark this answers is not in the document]
adrienbernede: Will do

uberenv.py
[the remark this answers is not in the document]
cyrush: @adrienbernede is it b/c you want to stream the output as well as parse it, as opposed to capture it all and print it?

I agree you will need to do that for such a long running command, however parsing the spack output of such a large operation might be brittle and a source of woes in the future. 

Can you run a separate command to check if it is already installed before the install command?  You might be able to get a more robust answer from a spack return code for a more specific command. If this is possible, then I don't think you need this change?

Overall your change to sexe looks ok - but I know that p.communicate()[0] is robust from years of use. I really don't want uberenv to deadlock on me :-) 

Sorry for so much guidance on what seems like such a minor change, but I want to apply caution.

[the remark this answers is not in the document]
adrienbernede: > @adrienbernede is it b/c you want to stream the output as well as parse it, as opposed to capture it all and print it?

Yes, that’s exactly what I meant.

> Can you run a separate command to check if it is already installed before the install command? You might be able to get a more robust answer from a spack return code for a more specific command. If this is possible, then I don't think you need this change?

It’s in my plan to make Uberenv more robust using the hash of the desired spec to guide subsequent spack commands. I want Uberenv to ask spack to concretize the desired spec, and then use the generated spec to check for path/already installed etc.
(e.g. `find_spack_pkg_path` isn’t robust as-is)

So once I have this improvement, I will be able to check in the spec is already installed, and won’t need this implementation of `sexe()` anymore.

But first, I want a reliable way to check my changes. It’s so easy to leave a bug in this process.

> Sorry for so much guidance on what seems like such a minor change, but I want to apply caution.

No worries, I’d rather have guidance now than later :)

adrienbernede:
LGTM

adrienbernede:
LGTM

uberenv.py
```diff
@@ -493,32 +522,51 @@ def install(self):
         # note: this assumes package extends python when +python
         # this may fail general cases
         if self.opts["install"] and "+python" in full_spec:
-            activate_cmd = "spack/bin/spack activate " + self.pkg_name
+            activate_cmd = "spack/bin/spack activate " + self.pkg_name + self.opts["spec"]
```
adrienbernede, line 525: Could use hash here: package already installed at this point, so hash is the robust way.

uberenv.py
```diff
@@ -493,32 +522,51 @@ def install(self):
...
             sexe(activate_cmd, echo=True)
         # if user opt'd for an install, we want to symlink the final
         # install to an easy place:
-        if self.opts["install"]:
-            pkg_path = self.find_spack_pkg_path(self.pkg_name)
+        if self.opts["install"] or "use_install" in self.opts:
+            pkg_path = self.find_spack_pkg_path(self.pkg_name,self.opts["spec"])
```
adrienbernede, line 530: Could use hash here: package already installed at this point, so hash is the robust way.

uberenv.py
```diff
@@ -493,32 +522,51 @@ def install(self):
...
+                if self.opts["install"]:
+                    pkg_lnk_dir = "{}-install".format(self.pkg_name)
+                    if os.path.islink(pkg_lnk_dir):
+                        os.unlink(pkg_lnk_dir)
+                    print("")
+                    print("[symlinking install to {}]".format(pjoin(self.dest_dir,pkg_lnk_dir)))
+                    os.symlink(pkg_path["path"],os.path.abspath(pkg_lnk_dir))
+                    print("")
+                    print("[install complete!]")
+        else:
```
adrienbernede, line 557: comment: `spack dev-build` case

adrienbernede:
@kennyweiss @cyrush 
This is open for review and suggestion.
I could document the CI addition. Changes in code should be almost transparent to usage, since that was the goal.

uberenv.py
```diff
@@ -455,30 +469,45 @@ def clean_build(self):
...
+        print(out)
+
+        for line in out.split("\n"):
+            if re.match(r"^(\[\+\]| - )  [a-z0-9]{32}  " + re.escape(self.pkg_name), line):
+                self.spec_hash = line.split("  ")[1]
+                if line.startswith("[+]"):
+                    pkg_path = self.find_spack_pkg_path_form_hash(self.pkg_name,self.spec_hash)
+                    install_path = pkg_path["path"]
+                    if os.path.isdir(install_path):
+                        print("[Warning: {} {} has already been install in {}]".format(self.pkg_name, self.opts["spec"],install_path))
```
adrienbernede, line 484: "installed"

uberenv.py
```diff
@@ -330,8 +333,16 @@ def setup_paths_and_dirs(self):
             sys.exit(-1)
 
 
-    def find_spack_pkg_path(self,pkg_name):
-        r,rout = sexe("spack/bin/spack find -p " + pkg_name,ret_output = True)
+    def find_spack_pkg_path_form_hash(self, pkg_name, pkg_hash):
```
kennyweiss, line 336: Typo: `_form_` -> `_from_`

uberenv.py
```diff
@@ -432,9 +444,11 @@ def patch(self):
         else:
             # let spack try to auto find compilers
             sexe("spack/bin/spack compiler find", echo=True)
-        dest_spack_pkgs = pjoin(spack_dir,"var","spack","repos","builtin","packages")
-        # hot-copy our packages into spack
-        sexe("cp -Rf {} {}".format(self.pkgs,dest_spack_pkgs))
+
+        if self.pkgs:
+            # hot-copy our packages into spack
```
kennyweiss, line 449: Suggestion/Nitpick: Maybe move the comment up one line and modify as follows:

```python
        # hot-copy our local packages into spack 
        if self.pkgs:
           dest_spack_pkgs = ...
```

Asking because I had to look up what ``self.pkgs`` contained.

uberenv.py
```diff
@@ -455,30 +469,45 @@ def clean_build(self):
...
 
     def show_info(self):
-        spec_cmd = "spack/bin/spack spec " + self.pkg_name + self.opts["spec"]
-        return sexe(spec_cmd, echo=True)
+        spec_cmd = "spack/bin/spack spec -IL " + self.pkg_name + self.opts["spec"]
+
+        res, out = sexe(spec_cmd, ret_output=True, echo=True)
+        print(out)
+
+        for line in out.split("\n"):
```
kennyweiss, line 477: Maybe add a comment above the `for` line to say what this is doing ?
Looks like it is looking for duplicates?

uberenv.py
```diff
@@ -455,30 +469,45 @@ def clean_build(self):
...
-        else:
-            install_cmd += "install "
-            if self.opts["run_tests"]:
-                install_cmd += "--test=root "
-        install_cmd += self.pkg_name + self.opts["spec"]
-        res = sexe(install_cmd, echo=True)
-        if res != 0:
-            return res
 
+        if not "use_install" in self.opts:
```
kennyweiss, line 494: Is the `use_install` option documented?

uberenv.py
```diff
@@ -493,32 +522,51 @@ def install(self):
...
-                if len(hcfg_glob) > 0:
-                    hcfg_path  = hcfg_glob[0]
-                    hcfg_fname = os.path.split(hcfg_path)[1]
-                    if os.path.islink(hcfg_fname):
-                        os.unlink(hcfg_fname)
-                    print("[symlinking host config file to {}]".format(pjoin(self.dest_dir,hcfg_fname)))
-                    os.symlink(hcfg_path,hcfg_fname)
-                print("")
-                print("[install complete!]")
+                # Symlink host-config file
```
kennyweiss, line 535: Thanks for the helpful comments!

.gitlab-ci.yml
```diff
@@ -0,0 +1,62 @@
...
+###############################################################################
+
+# Triggers Conduit pipeline asking for uberenv update
+# UPDATE_UBERENV variable, when set, will tell conduit to use the value as
+# a reference to the uberenv env revision to checkout on github.
+trigger-conduit:
+  variables:
+    UPDATE_UBERENV: $CI_COMMIT_REF_NAME
+  trigger:
+    project: radiuss/conduit
```
kennyweiss, line 52: If I understand this correctly, these trigger gitlab CI builds for conduit.

Will there be any feedback for the uberenv PR about the success of the gitlab task? E.g. block the PR if it fails?

uberenv.py
adrienbernede: Done

uberenv.py
adrienbernede: Done

uberenv.py
adrienbernede: For some reason I don’t remember, I added this as an option but only used and assignable internally, so not really an option.
I’m making is a `SpackEnv` field.

.gitlab-ci.yml
adrienbernede: Yes that’s what it is.

There is already a feedback, but since I used and abused of force push, it’s not obvious where to find it:

The last commit for which I added the "LGTM" sign-off has a testing status attached to it linking to the Gitlab corresponding CI pipeline.
https://github.com/LLNL/uberenv/commit/35a414505bc2d27b42b875cb36e021c738034024

This feedback could be used on Github side to prevent from merging. Of course, updating Uberenv in Serac and Conduit may not be straightforward for known reasons and the test would then fail.

uberenv.py
adrienbernede: Done

uberenv.py
adrienbernede: Done

uberenv.py
adrienbernede: Done

uberenv.py
adrienbernede: Done

uberenv.py
adrienbernede: Done

uberenv.py
adrienbernede: Done

adrienbernede:
LGTM

adrienbernede:
LGTM

adrienbernede:
@cyrush @kennyweiss @white238 May I merge the branch? I used to double review. :)

cyrush:
overall looks good. 

I checked out the gitlab outputs, are there some hidden errors in there?
` cp: cannot stat '/usr/WS2/bernede1/gitlab-runner/builds/zkQzXxmM/0/gitlab/radiuss/conduit/scripts/uberenv/packages/*': No such file or directory`

Are things getting applied as we expect?

adrienbernede:
LGTM

adrienbernede:
This is a good catch leading to an interesting point:
Uberenv CI is currently configured to trigger CI on a very specific branch of Conduit (an open PR), because Conduit do not have Gitlab CI in master. The branch I was pointing to was a bit outdated and did not contain any spack packages, hence the error, which has no critical consequence.

I keep force pushing micro changes to be able to trigger pipeline again.

PR #22, opened 2020-03-27
Adds Vcpkg as Uberenv's package manager on Windows

kennyweiss:
Upstreams changes made by @agcapps to Axom's version of ``uberenv`` for automated TPL generation on Windows using Vcpkg -- https://github.com/microsoft/vcpkg

Also updates comments and user docs to reflect the addition of Vcpkg.

See: https://github.com/LLNL/axom/blob/develop/scripts/uberenv for Axom's setup/usage.

### DIFF
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -1,5 +1,5 @@
 .. ############################################################################
-.. # Copyright (c) 2014-2018, Lawrence Livermore National Security, LLC.
+.. # Copyright (c) 2014-2020, Lawrence Livermore National Security, LLC.
 .. #
 .. # Produced at the Lawrence Livermore National Laboratory
 .. #
@@ -47,10 +47,12 @@
 Uberenv
 ~~~~~~~~~~~~~~~
 
-**Uberenv** automates using `Spack <ttp://www.spack.io>`_ to build and deploy software.
+**Uberenv** automates using a package manager to build and deploy software.
+It uses `Spack <http://www.spack.io>`_ on Unix-based systems (e.g. Linux and macOS)
+and `Vcpkg <https://github.com/microsoft/vcpkg>`_ on Windows systems.
 
-Many projects leverage `Spack <ttp://www.spack.io>`_ to help build the software dependencies needed to develop and deploy their projects on HPC systems. Uberenv is a python script that helps automate using Spack to build
-third-party dependencies for development and to deploy Spack packages.
+Many projects leverage package managers, like Spack and Vcpkg to help build the software dependencies needed to develop and deploy their projects on HPC systems. Uberenv is a python script that helps automate usage of a package manager to build
+third-party dependencies for development and deployment.
 
 Uberenv was released as part of Conduit (https://github.com/LLNL/conduit/). It is included in-source in several projects. The
 https://github.com/llnl/uberenv/ repo is used to hold the latest reference version of Uberenv.
@@ -59,15 +61,17 @@ https://github.com/llnl/uberenv/ repo is used to hold the latest reference versi
 uberenv.py
 ~~~~~~~~~~~~~~~~~~~~~
 
-``uberenv.py`` is a single file python script that automates fetching Spack, building and installing third party dependencies, and can optionally install packages as well.  To automate the full install process, ``uberenv.py`` uses a target Spack package along with extra settings such as Spack compiler and external third party package details for common HPC platforms.
+``uberenv.py`` is a single file python script that automates fetching Spack or Vcpkg, building and installing third party dependencies, and can optionally install packages as well.  To automate the full install process, ``uberenv.py`` uses a target Spack or Vcpkg package along with extra settings such as compilers and external third party package details for common HPC platforms.
 
-``uberenv.py`` is included directly in a project's source code repo in the folder: ``scripts/uberenv/``
-This folder is also used to store extra Spack and Uberenv configuration files unique to the target project. ``uberenv.py`` uses a ``project.json`` file to specify project details, including the target Spack package name and which Spack repo is used.  Conduit's source repo serves as an example for Uberenv and Spack configuration files, etc:
+``uberenv.py`` is typically included directly in a project's source code repo in the folder: ``scripts/uberenv/``
+This folder is also used to store extra configuration files unique to the target project. ``uberenv.py`` uses a ``project.json`` file to specify project details, including the target package name and the base branch or commit in the package manager.  
+
+Conduit's source repo serves as an example for Uberenv and Spack configuration files, etc:
 
 https://github.com/LLNL/conduit/tree/master/scripts/uberenv
 
 
-``uberenv.py`` is developed by LLNL in support of the `Ascent <http://github.com/alpine-dav/ascent/>`_, Axom, and `Conduit <https://github.com/llnl/conduit>`_  projects.
+``uberenv.py`` is developed by LLNL in support of the `Ascent <http://github.com/alpine-dav/ascent/>`_, `Axom <https://github.com/llnl/axom>`_, and `Conduit <https://github.com/llnl/conduit>`_  projects.
 
 
 Command Line Options
@@ -90,6 +94,7 @@ Build configuration
   ``--install``           Fully install target, not just dependencies    **False**
   ``--run_tests``         Invoke tests during build and against install  **False**
   ``--project-json``      File for project specific settings             ``project.json``
+  ``--triplet``           (vcpkg) Target architecture and linkage        ``x86-Windows``
  ======================= ============================================== ================================================
 
 The ``-k`` option exists for sites where SSL certificate interception undermines fetching
@@ -117,6 +122,14 @@ Default invocation on OSX:
                                       --spec %clang \
                                       --spack-config-dir scripts/uberenv/spack_configs/darwin/
 
+Default invocation on Windows:
+
+.. code:: bash
+
+    python scripts/uberenv/uberenv.py --prefix uberenv_libs \
+                                      --triplet x86-windows
+
+See `Vcpkg user docs <https://vcpkg.readthedocs.io/en/latest/users/triplets/>`_ for more information about triplets.
 
 Use the ``--install`` option to install the target package (not just its development dependencies):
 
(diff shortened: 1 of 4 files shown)


kennyweiss:
@white238 -- it looks like this might incorporate some of your changes from https://github.com/LLNL/uberenv/pull/19, which were already in Axom.

docs/sphinx/index.rst
```diff
@@ -90,6 +94,7 @@ Build configuration
   ``--install``           Fully install target, not just dependencies    **False**
   ``--run_tests``         Invoke tests during build and against install  **False**
   ``--project-json``      File for project specific settings             ``project.json``
+  ``--triplet``           (vcpkg) Target architecture and linkage        ``x86-Windows``
```
white238, line 97: Is triplet what vcpkg calls it? I find that a very generic name...

white238: Should have waited two seconds... i see you have a comment below

docs/sphinx/index.rst
```diff
@@ -162,11 +175,17 @@ Part of the configuration can also be addressed using a json file. By default, i
   package_version      **None**                   Spack package version                            **None**
   package_final_phase  ``--package-final-phase``  Controls after which phase Spack should stop     **None**
   package_source_dir   ``--package-source-dir``   Controls the source directory Spack should use   **None**
-  spack_url            **None**                   Url where to download Spack                      ``https://github.com/spack/spack.git``
+  spack_url            **None**                   Download url for Spack                          ``https://github.com/spack/spack.git``
```
white238, line 178: Might want to add vcpkg and spack specific tables for options...

uberenv.py
```diff
@@ -165,7 +172,7 @@ def parse_args():
     # option to force a spack pull
```
white238, line 172: Comment needs to be updated as well.

uberenv.py
```diff
@@ -233,28 +240,42 @@ def __init__(self, opts, extra_opts):
...
-    def set_from_args_or_json(self,setting):
-        try:
-            setting_value = self.project_opts[setting]
-        except (KeyError):
-            print("ERROR: {} must at least be defined in project.json".format(setting))
-            raise
+        self.dest_dir = os.path.abspath(self.opts["prefix"])
+        print("[installing to: {0}]".format(self.dest_dir))
+
+        # print a warning if the dest path already exists
```
white238, line 255: This is not entirely safe... if dest_dir is a file it will mkdir in the place and fail.

uberenv.py
```diff
@@ -269,6 +290,115 @@ def detect_platform(self):
...
+            print("[info: cloning vcpkg '{0}' branch from {1} into {2}]"
+                .format(vcpkg_branch,vcpkg_url, self.dest_vcpkg))
+
+            os.chdir(self.dest_dir)
+
+            clone_opts = ("-c http.sslVerify=false " 
+                          if self.opts["ignore_ssl_errors"] else "")
+
+            clone_cmd =  "git {0} clone -b {1} {2}".format(clone_opts, vcpkg_branch,vcpkg_url)
+            sexe(clone_cmd, echo=True)
```
white238, line 333: Check for failure on this command.

uberenv.py
```diff
@@ -269,6 +290,115 @@ def detect_platform(self):
...
+
+            # optionally, check out a specific commit
+            if "vcpkg_commit" in self.project_opts:
+                sha1 = self.project_opts["vcpkg_commit"]
+                print("[info: using vcpkg commit {0}]".format(sha1))
+                os.chdir(self.dest_vcpkg)
+                sexe("git checkout {0}".format(sha1),echo=True)
```
white238, line 340: Same.

uberenv.py
```diff
@@ -269,6 +290,115 @@ def detect_platform(self):
...
+                
+        if self.opts["repo_pull"]:
+            # do a pull to make sure we have the latest
+            os.chdir(self.dest_vcpkg)
+            sexe("git stash", echo=True)
```
white238, line 345: I know this is trying to be safe but are we expecting people to have changes? Shouldnt we error out in this case?

uberenv.py
```diff
@@ -269,6 +290,115 @@ def detect_platform(self):
...
+        res = sexe(install_cmd, echo=True)
+
+        # Running the install_cmd eventually generates the host config file,
+        # which we copy to the target directory.
+        src_hc = pjoin(self.dest_vcpkg, "installed", self.triplet, "include", self.pkg_name, "hc.cmake")
+        hcfg_fname = pjoin(self.dest_dir, "{0}.{1}.cmake".format(platform.uname()[1], self.triplet))
+        print("[info: copying host config file to {0}]".format(hcfg_fname))
+        shutil.copy(os.path.abspath(src_hc), hcfg_fname)
+        print("")
+        print("[install complete!]")
```
white238, line 398: Is it complete if it failed?

uberenv.py
```diff
@@ -233,28 +240,42 @@ def __init__(self, opts, extra_opts):
...
-        print("[uberenv options: {}]".format(str(self.opts)))
+        print("[uberenv project settings: {0}]".format(str(self.project_opts)))
+        print("[uberenv options: {0}]".format(str(self.opts)))
+
+        # setup main package name
+        self.pkg_name = self.project_opts["package_name"]
 
     def setup_paths_and_dirs(self):
-        self.uberenv_path = os.path.dirname(os.path.realpath(__file__))
+        self.uberenv_path = os.path.split(os.path.abspath(__file__))[0]
```
adrienbernede, line 250: I’m interested in the explanation behind this change. Was it causing an issue on Windows?
(I can’t remember why I chose realpath here, but I remember using it deliberately, should have commented...)

uberenv.py
```diff
@@ -464,10 +594,9 @@ def install(self):
         install_cmd = "spack/bin/spack "
         if self.opts["ignore_ssl_errors"]:
             install_cmd += "-k "
-        if not self.opts["install"]:
-            install_cmd += "dev-build -d {} -u {} ".format(self.pkg_src_dir,self.pkg_final_phase)
-        else:
-            install_cmd += "install "
+        install_cmd += "dev-build -d {0} ".format(self.pkg_src_dir)
+        if not self.opts["install"] and self.pkg_final_phase:
```
white238, line 598: This was brought in from Axom and has been decided to not be desired, though I would argue it will clean up some errors people will run into when your spack package doesn't have any phases (like bundle packages).

uberenv.py
```diff
@@ -464,10 +594,9 @@ def install(self):
...
+            install_cmd += "-u {0} ".format(self.pkg_final_phase)
```
adrienbernede, line 599: Again, I can’t remember why I wanted to keep `install` here, instead of dev-build.
I know that I think about it, it makes sense that one would want to install the development version. But I know there is a snag.

uberenv.py
adrienbernede: What was the reasoning exactly? I know we discussed it for #17, and maybe it was simply to preserve the original behavior.

uberenv.py
adrienbernede: Also, I think it’s best to specify the destination name. (In fact not done for spack clone neither).
I already ran into issues when trying to use a custom spack repo named `spack_<whatever>`... strange idea I know.

uberenv.py
adrienbernede: Since vcpkg is patched by Uberenv (without commit) I guess it makes sense?

white238: yes i keep forgetting that... i really wish we could get the spack guys to allow us a workflow where we don't have to patch spack..  ignore my comment.

uberenv.py
white238: So here is how I understand it and maybe we can reach an agreement.  We have two cases here:

1) Normal case (for uberenv at least):

We have a package which is a full package (like axom) where we want "uberenv axom" to install just the dependencies and stop at the final phase.  Then "uberenv --install axom" to install deps and axom itself.

2) My new introduced case which breaks the devtools away from main package. Because it is a spack bundlepackage it doesn't have phases. So i wanted "uberenv axomdevtools" to be the same as "uberenv --install axomdevtools" because there isn't a final phase to define.

adrienbernede: You could define a fake phase. (I know, what an elegant solution ...)

adrienbernede: I think the problem was that if final phase is undefined, then Uberenv defaults to install the package and not only it’s dependencies. And that would be a change of the workflow.

white238:
Moved my small changes from #19 into this branch so we avoid a large set of conflicts.

uberenv.py
```diff
@@ -464,10 +594,9 @@ def install(self):
...
             if self.opts["run_tests"]:
```
white238, line 599: While i was reverting some work, I noticed this run test option was not in the "not install" if.  Should this be deindented so it hits both cases?

adrienbernede: Thanks for noticing this, now I remember!
So, in #20 I noticed that there were no more testing support because dev-build does not support tests. That’s why in re-introduced the `install` command.
This also explain why we cannot "deindent" this.
Maybe we could ask the Spack guys to add support for tests in the dev-build command as-well.

uberenv.py
cyrush: if folks have changes, they should be in their copy of the packages or config files (which will get copied over after this cleans up). Doing git stash was the only sane way to clean things up and make it so stuff doesn't turned into a tangled mess run-to-run.

uberenv.py
cyrush: i think this command can also fail if there is already a clone, in that case we want to move forward ... 

Making this bullet proof is hard, but making things smooth for repeated calls (given that downloads aren't robust)  is super important.

uberenv.py
adrienbernede: (or `git reset --hard HEAD` if you don’t want to keep the changes in the stash list)

cyrush: I believe we need `git stash` b/c the `spack_commit` option can leave is in a detached head state, confusing `reset HEAD`

adrienbernede: That’s indeed likely to happen. If it were me, I would always create a branch (e.g. with the `date`) in order to avoid any detached head magic.

adrienbernede: But you could end up with many branches at the same spot... Or create a branch with a unique uberenv-specific name.

adrienbernede: OK, OK, let’s keep things simple !?

uberenv.py
kennyweiss: I am not sure that it was intentional.
Crawling through the git history, it might have been done by git automerge?

I'll change it back, but I'm curious about the difference.

It seems to me like your change would be better able to handle things like windows directories with spaces in them (?)

kennyweiss:
Superseded by #40

PR #25, opened 2020-06-04
Fix: output and spack commit

adrienbernede:
Five things:

- Fix missing outputs: 
    A mistake in output management caused all spack outputs in spack dev-build to be ignored.
    Spack dep-build remains in "quiet" mode, avoiding outputs for successful builds.
- Better spack commit:
    When referring to a specific Spack commit, Uberenv would not change the current version of Spack and not print an error.
    I chose to force the update if the current Spack is not at the specified commit, without removing existing installations.
- Variable naming:
    There were varations in the name of the variables used to retrieve the outputs of sexe.
    I may it more consistent for the sake of readability.
- Return value on failure.
- Add info output for patched packages.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -511,7 +511,10 @@ class SpackEnv(UberEnv):
                 if self.opts["run_tests"]:
                     install_cmd += "--test=root "
             install_cmd += self.pkg_name + self.opts["spec"]
-            res, out = sexe(install_cmd, ret_output=True, echo=True)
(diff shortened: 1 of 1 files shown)


adrienbernede:
LGTM

adrienbernede:
@white238 @kennyweiss @cyrush 
These are simple bug fixes and minor changes I use in all my instances.
Can I merge it? Could someone review it?

cyrush:
looks good, feel free to merge

PR #28, opened 2020-07-16
Fix mirror issue

### LINKED ISSUE #27: mirror does not respect passed spec
need to change:

https://github.com/LLNL/uberenv/blob/master/uberenv.py#L608

to include the full spec, not the pkg name

adrienbernede:
Fixes #27


adrienbernede:
LGTM

adrienbernede:
CI jobs failed due to time limit. We have a look at it, but they pass when relaunched. So it’s good to me.

PR #34, opened 2020-11-13
Allow to use uberenv to only setup spack

adrienbernede:
This comes from my current work on MFEM CI, but could benefit others:

In CI, it is valuable to cache `uberenv_libs` at its original state (spack downloaded and configured).
That way we can skip spack download in all the jobs (and replace it with a local cache download).

For more about setting up such a cache, see: 
https://github.com/mfem/mfem/blob/5a4025dc841baab5f85d33d0febbf2d20ba465bb/.gitlab/quartz.yml#L25
https://github.com/mfem/mfem/blob/5a4025dc841baab5f85d33d0febbf2d20ba465bb/.gitlab/quartz.yml#L35

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -195,12 +195,12 @@ def parse_args():
                            " but can cause issues with macOS versions >= 10.13. "
                            " so it is disabled by default.")
 
-    # option to stop after spack config
-    parser.add_option("--spack-only",
+    # option to stop after spack download and setup
+    parser.add_option("--setup-only",
                       action="store_true",
-                      dest="spack_only",
+                      dest="setup_only",
                       default=False,
-                      help="Stops uberenv right after spack config. Spec has no effect")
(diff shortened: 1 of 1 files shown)


adrienbernede:
@white238 I think this is a harmless change: If you agree, I merge it right away ;)

white238 (approved):
Yepp I can see the benefit of this! approved!

uberenv.py
```diff
@@ -195,6 +195,13 @@ def parse_args():
                            " but can cause issues with macOS versions >= 10.13. "
                            " so it is disabled by default.")
 
+    # option to stop after spack config
+    parser.add_option("--spack-only",
```
white238, line 199: Though maybe a different name for this, though the help comment helps.  Maybe clarifying that to be:

Download and configure Spack only.  No further spack commands will be called.

adrienbernede: Improved!

Issue #31, opened 2020-09-02
Uberenv as submodule


adrienbernede:
Uberenv is being used in more and more projects, especially for CI purpose.

However, the fact that is has to be added in-source is a brake to broader adoption and ease of use.

Adapting uberenv to be usable as a submodule will probably:
- break the current workflow
- require multiple small changes

Still, I think this should be done. If we do it in a separate branch at first, we may have trouble migrating to Spack environments afterward (duplicate the amount of work).

So I’m considering completing my work on environments first, even if some changes in Spack will likely imply adapting uberenv later on, hopefully drastically reducing its scope.

Input welcome

adrienbernede:
@white238 @kennyweiss @cyrush, any thought about this?
I’ll create an issue to discuss the impact of Gregs PR (https://github.com/spack/spack/pull/15256).

white238:
I don't see how changing uberenv to a submodule makes it easier to upgrade.  They are both essentially locked into a SHA.  The only difference is a copy vs a "cd uberenvSubmodule; git checkout develop; git pull".  The nice thing about having it directly in the repo is I can put the script (and only the script) exactly where I want it in the repo.

adrienbernede:
Locked into a SHA?

Copying uberenv in a project implies losing the history, and any information about what version of Uberenv we are using and how to merge in new changes.

I agree that the submodule will apply only to the script (and possibly to some shared spack configurations, like compilers for machines) which limits the interest of the process.

OK, I’ll try to come back better prepared for this battle :)

white238:
By locking into a SHA, I meant locking into a point in history.  Yes i lose history one way but as a user of uberenv I don't really care about the history that much.  If i do care I can go look at the uberenv repo.

Also if i remember correctly uberenv requires the spack_configs and packages directories to at the same level as the script.  This could obviously be solved with a simple change in uberenv.

adrienbernede:
Yes, that’s part of the changes that would need to be done.

I’m going back and forth with this submodule thing. Clearly, some projects don’t like adding another project, in source, and I understand them.

"Uberenv is just a script" and for a long while it hasn’t be changing much. But during the last year, it has changed, and I have it in at least 6 projects.

cyrush:
@adrienbernede you can use uberenv a a submodule in any project you like, but its a per project decision. 

We discussed this for axom, we didn't see a strong benefit, as we would also have to change the directory structure and how it was called if it was a submodule.

[...]

PR #35, opened 2020-11-26
Introduce minimum changes to get uberenv usable in submodule

adrienbernede:
This addresses #31 

The strategy here was to introduce as few changes as possible.
In fact only a few changes were required:

- looking for `.uberenv/config.json` in parent directories if `project.json` is not found locally.
- defining uberenv_conf_path as the parent directory of the configuration file, and retrieve `spack_configs` and `packages` from there.

As a result, Uberenv works as before in the following configuration:

Project
|_ .uberenv
&ensp;|_ config.json
&ensp;|_ spack_configs
&ensp;|_ packages
|_ scripts
&ensp;|_ uberenv
&ensp;&ensp;|_ uberenv.py

### DIFF
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -135,7 +135,7 @@ If the target Spack package supports Spack's testing hooks, you can run tests du
 For details on Spack's spec syntax, see the `Spack Specs & dependencies <http://spack.readthedocs.io/en/latest/basic_usage.html#specs-dependencies>`_ documentation.
 
 
-Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_config/{platform}`` or you can use the **--spack-config-dir** option to specify a directory with compiler and packages yaml files to use with Spack. See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_
+Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_configs/{platform}`` or under ``.uberenv/spack_configs/{platform}``. You may also use the **--spack-config-dir** option to specify a directory with compiler and packages yaml files to use with Spack. See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_
 and `Spack System Packages
 <http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>`_
 documentation for details.
@@ -167,6 +167,9 @@ Part of the configuration can also be addressed using a json file. By default, i
   spack_activate       **None**                   Spack packages to activate                       **None**
  ==================== ========================== ================================================ =======================================
 
+However, Uberenv config can also be setup to sit in a location external to the uberenv directory itself. In particular, if no ``project.json`` file is found there, the script will look for ``.uberenv/config.json`` recursively in parent directories. The typical usage is to place it at the root directory of the project.
+
+When ``.uberenv/config.json`` is found, it is assumed that ``.uberenv`` also contains the ``packages`` and the ``spack_configs`` directories.
 
 Optimization
 ------------
@@ -187,7 +190,7 @@ Project Settings
 
 A few notes on using ``uberenv.py`` in a new project:
 
-* For an example of how to craft a ``project.json`` file a target project, see: `Conduit's project.json file <https://github.com/LLNL/conduit/tree/master/scripts/uberenv/project.json>`_
+* For an example of how to craft a ``project.json`` / ``.uberenv/config.json`` file a target project, see: `Conduit's project.json file <https://github.com/LLNL/conduit/tree/master/scripts/uberenv/project.json>`_
 
 * ``uberenv.py`` hot copies ``packages`` to the cloned Spack install, this allows you to easily version control any Spack package overrides necessary
 
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -141,7 +141,7 @@ Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_configs
 * ``{spack_configs_path}`` can be specified in the json config file.
 
 You may instead use the **--spack-config-dir** option to enforce the use of a specific directory. As long as it provides Uberenv with the yaml files to use with Spack.
-See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_ and `Spack System Packages <http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>` documentation for details.
+See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_ and `Spack System Packages <http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>`_ documentation for details.
 
 .. note::
     The bootstrapping process ignores ``~/.spack/compilers.yaml`` to avoid conflicts
@@ -171,7 +171,7 @@ Part of the configuration can also be addressed using a json file. By default, i
   spack_activate       **None**                   Spack packages to activate                       **None**
  ==================== ========================== ================================================ =======================================
 
-However, Uberenv config can also be setup to sit in a location external to the uberenv directory itself. In particular when Uberenv is used as a submodule. Uberenv identifies such a case when no ``project.json`` file is found next to the script. It will then look for ``.uberenv_config.json`` recursively in parent directories. The typical usage is to place it at the root directory of the project.
+However, Uberenv configuration may instead sit in a location external to the uberenv directory itself. In particular when Uberenv is used as a submodule. Uberenv identifies such a case when no ``project.json`` file is found next to the script. It will then look for ``.uberenv_config.json`` recursively in parent directories. The typical usage is to place it at the root directory of the project.
 
 When used as a submodule ``.uberenv_config.json`` should define both ``spack_configs_path`` and ``spack_packages_path``, providing Uberenv with the respective location of ``spack_configs`` and ``packages`` directories. Indeed, they cannot sit next to ``uberenv.py`` as per default, since Uberenv repo does not provide them.
 
@@ -197,5 +197,4 @@ A few notes on using ``uberenv.py`` in a new project:
 * For an example of how to craft a ``project.json`` / ``.uberenv_config.json`` file a target project, see: `Axom's project.json file <https://github.com/LLNL/axom/tree/develop/scripts/uberenv/project.json>`_
 
 * ``uberenv.py`` hot copies ``packages`` to the cloned Spack install, this allows you to easily version control any Spack package overrides necessary
-
 
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -45,7 +45,7 @@
 .. _building_with_uberenv:
 
 Uberenv
-~~~~~~~~~~~~~~~
+~~~~~~~
 
 **Uberenv** automates using `Spack <ttp://www.spack.io>`_ to build and deploy software.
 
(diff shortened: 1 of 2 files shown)


adrienbernede:
@white238 I should have mentioned you on this.
Does it looks like what you were expecting?

docs/sphinx/index.rst
```diff
@@ -135,7 +135,7 @@ If the target Spack package supports Spack's testing hooks, you can run tests du
 For details on Spack's spec syntax, see the `Spack Specs & dependencies <http://spack.readthedocs.io/en/latest/basic_usage.html#specs-dependencies>`_ documentation.
 
 
-Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_config/{platform}`` or you can use the **--spack-config-dir** option to specify a directory with compiler and packages yaml files to use with Spack. See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_
+Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_configs/{platform}`` or under ``.uberenv/spack_configs/{platform}``. You may also use the **--spack-config-dir** option to specify a directory with compiler and packages yaml files to use with Spack. See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_
```
kennyweiss, line 138: Is this a typo fix? Or a change from `scripts/uberenv/spack_config` to `scripts/uberenv/spack_configs` ?

kennyweiss (approved):
Looks good to me @adrienbernede.  

Could you please confirm that you've checked this against an existing usercode (e.g. axom or conduit) and that this update does not require any changes to those projects?

white238:
> Looks good to me @adrienbernede.
> 
> Could you please confirm that you've checked this against an existing usercode (e.g. axom or conduit) and that this update does not require any changes to those projects?

This would require projects to move their directory.  I was currently talking to @adrienbernede about this and probably should have had the converstation here.

Originally I was thinking that this would be a config file not a directory. Something like:

.uberenv:

```
spack_configs = path/in/repo/to/config/dir
spack_packages = path/in/repo/to/packages/dir

# Then maybe basic options
ForcePrefixOnCommandline=True
```

My concern with this approach is "hiding" files/directories in a hidden folder that users/developers need to regularly change. This is possibly unfounded but I find mildly confusing but @adrienbernede made a good point that it would have all projects act the same and people wouldn't need to run around looking for these files/directories in different projects.

docs/sphinx/index.rst
adrienbernede: Typo fix.

adrienbernede:
> Looks good to me @adrienbernede.
> 
> Could you please confirm that you've checked this against an existing usercode (e.g. axom or conduit) and that this update does not require any changes to those projects?

Checked against Umpire. Not conduit or axom, will do.

adrienbernede:
> > Looks good to me @adrienbernede.
> > Could you please confirm that you've checked this against an existing usercode (e.g. axom or conduit) and that this update does not require any changes to those projects?
> 
> This would require projects to move their directory. I was currently talking to @adrienbernede about this and probably should have had the converstation here.
> 
> Originally I was thinking that this would be a config file not a directory. Something like:
> 
> .uberenv:
> 
> ```
> spack_configs = path/in/repo/to/config/dir
> spack_packages = path/in/repo/to/packages/dir
> 
> # Then maybe basic options
> ForcePrefixOnCommandline=True
> ```
> 
> My concern with this approach is "hiding" files/directories in a hidden folder that users/developers need to regularly change. This is possibly unfounded but I find mildly confusing but @adrienbernede made a good point that it would have all projects act the same and people wouldn't need to run around looking for these files/directories in different projects.

@white238 There should be no required change for the "legacy" usage: the goal is to first look in the usual location, and otherwise "switch" to looking for an external config (.uberenv in parent dir).

white238:
> > > Looks good to me @adrienbernede.
> > > Could you please confirm that you've checked this against an existing usercode (e.g. axom or conduit) and that this update does not require any changes to those projects?
> > 
> > 
> > This would require projects to move their directory. I was currently talking to @adrienbernede about this and probably should have had the converstation here.
> > Originally I was thinking that this would be a config file not a directory. Something like:
> > .uberenv:
> > ```
> > spack_configs = path/in/repo/to/config/dir
> > spack_packages = path/in/repo/to/packages/dir
> > 
> > # Then maybe basic options
> > ForcePrefixOnCommandline=True
> > ```
> > 
> > 
> > My concern with this approach is "hiding" files/directories in a hidden folder that users/developers need to regularly change. This is possibly unfounded but I find mildly confusing but @adrienbernede made a good point that it would have all projects act the same and people wouldn't need to run around looking for these files/directories in different projects.
> 
> @white238 There should be no required change for the "legacy" usage: the goal is to first look in the usual location, and otherwise "switch" to looking for an external config (.uberenv in parent dir).

The problem is having a file/directory you need to edit on a normal basis being hidden.  This directory wouldn't even show up with a normal `ls` command.

adrienbernede:
I hear you @white238. But @kennyweiss was asking if this implied changes for projects like Axom and Conduit, and you basically said yes. I was answering to that: no change require.

I agree that you may disagree (:P) on the way the external config is implemented. I followed your git analogy all the way to making uberenv config a hidden directory.

I will add configurable `spack-configs` and `packages` paths.

Still, we need a default location, in my opinion, and I liked that it could be in this hidden directory, since we already require a hidden json config file.

So here is my proposal:

Legacy mode:
"Uberenv script dir also contains project.json/packages/spack-configs"
Submodule mode:
"Uberenv config.json location (I shouldn't have renamed it I'm thinking) determine the default location of packages and spack-configs"
New option:
"packages and spack-configs path can be configured in the json file"

If you agree, what is left to determine is if we really don’t want a `.uberenv` directory. I like it because it gives a clean default location. What do you think @white238.

kennyweiss:
@adrienbernede  -- Thank you for checking that the current approach does not require any changes for our current users.

After reading through the discussion again, I like @white238's strategy of having a top-level `.uberenv` or `.uverenv_config` _file_  that can point to the proper locations, rather than a hidden folder akin to `.git`.  That way user codes are free to choose where the submodule lives and where the spack_config files (and vcpkg config files) live within their repo.

adrienbernede:
@kennyweiss So you would like to make `spack_configs` and `packages` locations a required setting, right? But only in submodule mode I guess.

adrienbernede:
Or should the default location still be where the uberenv.py script lives? That may be the most straightforward solution.

kennyweiss:
@adrienbernede -- I would think the `.uberenv` file could be used to explicitly set the paths, but the default would be where `uberenv.py` lives. That would given backward compatibility / simpler integration with the existing users.

cyrush:
my 2 cents: 
I really don't care about the file being hidden vs not hidden. 

I suggest the contents stay json, vs some other format. (yaml doesn't even work out of the box in some python installs)

I guess I would vote for:
`uberenv_config.json`


I still don't understand how it would be located, one dir up from the uberenv submodule?

cyrush:
another q:
The implicit SYS_TYPE case and darwin support, etc  are what drives the spack-configs dir right? 

In other cases, they are just passed.

adrienbernede:
@cyrush uberenv.py scours for it in the parent directories until its found or reach the root dir.

cyrush:
> @cyrush uberenv.py scours for it in the parent directories until its found or reach the root dir.

got it, like git does.

adrienbernede:
Answer to other q:
The first (new) step is to locate "spack_configs", then $SYS_TYPE or darwin is used to deduce which sub-dir to use.

adrienbernede:
@cyrush actually, I understand why this other question.
In fact, I chose to have a default location for `spack_configs` not to conflict with the location of `spack_configs/<arch>` which you can enforce on command line.
With @kennyweiss and @white238 requirements, I will probably have deal with this, like having a `--spack-config-path` option and `--spack-config` option, the latest being the legacy one.

adrienbernede:
In fact I will leave `spack-config-path` as a setting in the json file, and keep a single option in uberenv to enforce a specific configuration.

white238:
> In fact I will leave `spack-config-path` as a setting in the json file only, and keep a single option in uberenv to enforce a specific configuration.

This should probably be `spack-configs-path`, i feel the plural is more clear.

adrienbernede:
Almost there: tested on Umpire and on Axom (on quartz: `python scripts/uberenv/uberenv.py` passed).
Need to update documentation.

docs/sphinx/index.rst
```diff
@@ -135,10 +136,12 @@ If the target Spack package supports Spack's testing hooks, you can run tests du
...
-and `Spack System Packages
-<http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>`_
-documentation for details.
+Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_configs/{platform}`` or under ``{spack_config_paths}/{platform}``, where:
+* ``{platform}`` must match the platform determined by uberenv.
+* ``{spack_configs_path}`` can be specified in the json config file.
+
+You may instead use the **--spack-config-dir** option to enforce the use of a specific directory. As long as it provides Uberenv with the yaml files to use with Spack.
+See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_ and `Spack System Packages <http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>` documentation for details.
```
adrienbernede, line 144:
```suggestion
See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_ and `Spack System Packages <http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>`_ documentation for details.
```

docs/sphinx/index.rst
```diff
@@ -167,6 +171,9 @@ Part of the configuration can also be addressed using a json file. By default, i
   spack_activate       **None**                   Spack packages to activate                       **None**
  ==================== ========================== ================================================ =======================================
 
+However, Uberenv config can also be setup to sit in a location external to the uberenv directory itself. In particular when Uberenv is used as a submodule. Uberenv identifies such a case when no ``project.json`` file is found next to the script. It will then look for ``.uberenv_config.json`` recursively in parent directories. The typical usage is to place it at the root directory of the project.
```
adrienbernede, line 174:
```suggestion
However, Uberenv configuration may instead sit in a location external to the uberenv directory itself. In particular when Uberenv is used as a submodule. Uberenv identifies such a case when no ``project.json`` file is found next to the script. It will then look for ``.uberenv_config.json`` recursively in parent directories. The typical usage is to place it at the root directory of the project.
```

uberenv.py
```diff
@@ -221,13 +222,17 @@ def parse_args():
...
 
 
 def uberenv_script_dir():
     # returns the directory of the uberenv.py script
-    return os.path.dirname(os.path.abspath(__file__))
+    return os.path.dirname(os.path.realpath(__file__))
+
+def uberenv_config_dir(project_json_file):
+    # returns the directory of the uberenv.py script
+    return os.path.dirname(os.path.realpath(project_json_file))
```
adrienbernede, line 235:
```suggestion
```

uberenv.py
```diff
@@ -252,7 +278,8 @@ def __init__(self, opts, extra_opts):
         print("[uberenv options: {}]".format(str(self.opts)))
 
     def setup_paths_and_dirs(self):
-        self.uberenv_path = os.path.dirname(os.path.realpath(__file__))
+        # Script dir
+        self.uberenv_path = uberenv_script_dir()
```
adrienbernede, line 282:
```suggestion
        self.uberenv_path = uberenv_script_dir()
```

adrienbernede (commented):
Auto review: some fixes.

white238 (approved):
Thanks @adrienbernede I look forward to using this in our projects.

Issue #32, opened 2020-10-16
allow mirror url to be set in project.json file


cyrush:
it makes life easier to have an option to stored it here, vs always pass on the command line

adrienbernede:
@cyrush what about creating a `mirrors.yaml` file in the desired `spack_configs/<sys_type>` and modify uberenv so that it copies it as well? (Just tested, it works fine.)

Note: with environments this won't be necessary anymore (#18).

cyrush:
that does seem like a good option, but I am using  the same mirror across configs, 
creating lots of mirror.yaml files its harder to manage vs setting in one place.

white238:
We really need to move to environments. They solve a lot of these problems.

adrienbernede:
We could apply the same logic as for "config.yaml": copy the top level one, and if present override with the one in the spack_config dir.

adrienbernede:
@white238 agreed. PR ready, and demo in Umpire OK ;)

cyrush:
in that we can cascade them (nest scopes depending on platform)?

adrienbernede:
> in that we can cascade them (nest scopes depending on platform)?

Sorry, not sure what you are asking.

cyrush:
My question was for @white238, so to how env's help. 
I think env's help b/c we can have overall settings (like mirror settings) and then include another env file with packages and compilers?

@adrienbernede  can you point me to the umpire stuff?

Already down a rabbit hole for a spack update, I am planning on doing an uberenv update as well. 
I have env files I can test.

adrienbernede:
Umpire with "Environment based version of Uberenv":
https://github.com/LLNL/Umpire/pull/481

white238:
Just that we keep running into problems that environments solve.  I thought you could add mirrors to environment files but i can't seem to find it in their documentation.

adrienbernede:
@white238 You just can.

adrienbernede:
Syntax is just like if you included the mirrors.yaml file.

PR #37, opened 2020-12-04
Make mirror configurable in spack_config

adrienbernede:
Addresses #32 

This is a basic proposal to address the issue.
However it is not protected against some cases, like if someone configures a mirror in the json file, and then another on command line.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -443,17 +443,23 @@ class SpackEnv(UberEnv):
         # copy in "defaults" config.yaml
         config_yaml = os.path.abspath(pjoin(self.uberenv_path,"spack_configs","config.yaml"))
         sexe("cp {} {}/".format(config_yaml, spack_etc_defaults_dir ), echo=True)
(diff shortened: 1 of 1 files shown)


white238:
Unless I'm missing something, having the `mirror.yaml` not under the `SYS_TYPE` folder seems like it wouldn't work.  For example, if I put `/usr/workspace/some_shared_space/mirror` like we normally do for LC, what happens on your personal machine?  Is Spack have logic in there to toggle based on environment variables?  Reading their documentation isn't helping me find it.

cyrush:
@white238 in the wonderful universe of possibilities - we have https mirrors that work anywhere the internets work.

white238:
Can you list mirrors and it just keeps going down the list if it can't access them?

adrienbernede:
Of course we can do something more advanced. I was trying to adjust the effort to the minimum requirement.
In general, the way we handle the configuration lack some uniformity and robustness.
We could rely more on Spack commands now that they exists (like the one you showed me yesterday @white238).

white238 (approved):

Issue #36, opened 2020-12-02
Add a RELEASE_NOTES file to better track changes


kennyweiss:
The format should follow [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)

See Axom for an example: https://github.com/LLNL/axom/blob/develop/RELEASE-NOTES.md

PR #39, opened 2020-12-14
Various improvements needed for Serac

white238:
* Add ability to have multiple packages directories
* Add ability to force command line prefix
* Pretty print options to screen
* Added RELEASE-NOTES.md #36

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -14,7 +14,8 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
 
 ### Changed
 - Added ability to have multiple packages directories that will get copied into spack on top of
-  each other via renamed project option: `spack_packages_path` to `spack_packages_paths`
+  each other via project configuration option: `spack_packages_path`
 - Pretty print various options to screen for readability
+- Allow `.uberenv_config.json` to live at the same level as `uberenv.py`
 
 ### Fixed
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,0 +1,20 @@
+# Uberenv Software Release Notes
+
+Notes describing significant changes in each Uberenv release are documented
+in this file.
+
+The format of this file is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
+
+The Uberenv project release numbers follow [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+### Added
+- Allow projects to force prefix on command line via new project option: `force_commandline_prefix`
+
+### Changed
+- Added ability to have multiple packages directories that will get copied into spack on top of
+  each other via renamed project option: `spack_packages_path` to `spack_packages_base_paths`
+- Pretty print various options to screen for readability
+
+### Fixed
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -20,5 +20,6 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
 - Pretty print various options to screen for readability
 - Allow `.uberenv_config.json` to live at the same level as `uberenv.py`
 - No longer removes symlinks when using the directory of `uberenv.py`
+- Reduce Spack's git history to a bare minimum
 
 ### Fixed
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -14,7 +14,7 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
 
 ### Changed
 - Added ability to have multiple packages directories that will get copied into spack on top of
-  each other via renamed project option: `spack_packages_path` to `spack_packages_base_paths`
+  each other via renamed project option: `spack_packages_path` to `spack_packages_paths`
 - Pretty print various options to screen for readability
 
 ### Fixed
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -10,7 +10,9 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
 ## Unreleased
 
 ### Added
-- Allow projects to force prefix on command line via new project option: `force_commandline_prefix`
+- Allow projects to force specifying `--prefix` on command line via new project option:
+  `force_commandline_prefix`.  This is useful when the default `uberenv_libs` can not exist
+  inside of your project's source repository.
 
 ### Changed
 - Added ability to have multiple packages directories that will get copied into spack on top of
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -19,5 +19,6 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
   each other via project configuration option: `spack_packages_path`
 - Pretty print various options to screen for readability
 - Allow `.uberenv_config.json` to live at the same level as `uberenv.py`
+- No longer removes symlinks when using the directory of `uberenv.py`
 
 ### Fixed
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -47,9 +47,9 @@
 Uberenv
 ~~~~~~~
 
-**Uberenv** automates using `Spack <ttp://www.spack.io>`_ to build and deploy software.
+**Uberenv** automates using `Spack <https://spack.io>`_ to build and deploy software.
 
-Many projects leverage `Spack <ttp://www.spack.io>`_ to help build the software dependencies needed to
+Many projects leverage `Spack <https://spack.io>`_ to help build the software dependencies needed to
 develop and deploy their projects on HPC systems. Uberenv is a python script that helps automate using Spack to build
 third-party dependencies for development and to deploy Spack packages.
 
@@ -75,7 +75,7 @@ Uberenv can also be used as a submodule. In that case, it is required to provide
 ``.uberenv_config.json`` in a parent directory. This file is similar to ``project.json`` in purpose, but should
 additionally provide the entries ``spack_configs_path`` and ``spack_packages_path``. Details in :ref:`project_configuration`.
 
-Uberenv is developed by LLNL originally in support of the `Ascent <http://github.com/alpine-dav/ascent/>`_,
+Uberenv is developed by LLNL originally in support of the `Ascent <https://github.com/alpine-dav/ascent/>`_,
 `Axom <https://github.com/llnl/axom>`_, and `Conduit <https://github.com/llnl/conduit>`_  projects. It is now also used
 in `Umpire <https://github.com/llnl/umpire>`_, `CHAI <https://github.com/llnl/CHAI>`_, `RAJA <https://github.com/llnl/RAJA>`_
 and `Serac <https://github.com/llnl/serac>`_ for example.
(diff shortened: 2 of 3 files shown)


uberenv.py
```diff
@@ -361,16 +375,26 @@ def setup_paths_and_dirs(self):
...
-            new_path = self.project_opts["spack_packages_path"]
-            if new_path is not None:
-                spack_packages_path = pabs(new_path)
-        # All local packages
-        self.pkgs = pjoin(spack_packages_path, "*")
+        # Find project level packages to override spack's internal packages
+        self.spack_packages_base_paths = []
+        if "spack_packages_base_paths" in self.project_opts.keys():
+            # packages directories listed in project.json
+            _paths = self.project_opts["spack_packages_base_paths"]
```
white238, line 382: Not sold on this option name but wanted to show that it wasn't the individual packages directory ("packages" vs "packages/axom, packages/mfem, etc...")

docs/sphinx/index.rst
```diff
@@ -154,26 +154,36 @@ copied to destination directory. This file specifies the compiler settings and p
...
+  package_final_phase       ``--package-final-phase``  Controls after which phase Spack should stop     **None**
+  package_source_dir        ``--package-source-dir``   Controls the source directory Spack should use   **None**
+  spack_url                 **None**                   Url where to download Spack                      ``https://github.com/spack/spack.git``
+  spack_commit              **None**                   Spack commit to checkout                         **None**
+  spack_activate            **None**                   Spack packages to activate                       **None**
+  spack_packages_base_paths **None**                   Spack packages to be copied over cloned Spack    ``packages``
+ ========================== ========================== ================================================ =======================================
+
+When used as a submodule ``.uberenv_config.json`` should define both ``spack_configs_path`` and ``spack_packages_path``, providing Uberenv with the respective location of ``spack_configs`` and ``packages`` directories. Indeed, they cannot sit next to ``uberenv.py`` as per default, since the Uberenv repo does not provide them.
```
adrienbernede, line 182: spack_packages_path -> spack_packages_base_paths

uberenv.py
adrienbernede: I relied on the plural form:
spack_packages_path for all packages,
spack_package_path for a specific package.

uberenv.py
```diff
@@ -361,16 +375,26 @@ def setup_paths_and_dirs(self):
...
 
         # setup destination paths
+        if not self.opts["prefix"]:
+            if self.project_opts["force_commandline_prefix"]:
+                # project has specified prefix must be on command line
+                print("[ERROR: --prefix flag for library destination is required]")
+                sys.exit(1)
+            # otherwise set default
+            self.opts["prefix"] = "uberenv_libs"
+
```
adrienbernede, line 397: I understand the usage.
A general note: options in Uberenv now tend to have diverging treatment. One has a default value, the other is expected to be set in the project.json, another can be forced to be defined on command line, etc.

uberenv.py
```diff
@@ -511,10 +535,13 @@ def patch(self):
             sexe("spack/bin/spack compiler find", echo=True)
 
         # hot-copy our packages into spack
-        if self.pkgs:
+        if self.spack_packages_base_paths:
             dest_spack_pkgs = pjoin(spack_dir,"var","spack","repos","builtin","packages")
-            print("[copying patched packages from {0}]".format(self.pkgs))
-            sexe("cp -Rf {} {}".format(self.pkgs,dest_spack_pkgs))
+            for _base_path in self.spack_packages_base_paths:
```
adrienbernede, line 540: The order in the dictionary is quite important, right? Maybe we should mention it in the doc?

adrienbernede (commented):
Looks good at first look. I would like to do a second pass, and test it.

uberenv.py
white238: > I relied on the plural form:
> spack_packages_path for all packages,
> spack_package_path for a specific package.

so `spack_packages_paths`? since its the packages directory and there are multiples of them? This already confuses me and I did it.

adrienbernede: That works for me. With the weakness that every "s" counts.

uberenv.py
white238: Yes ... we should probably do a once over and conform many things in uberenv.  This was the cleanest way, I felt, to check for a non-user specified path.

uberenv.py
white238: spack_packages_path_list?

adrienbernede: I don't have strong opinions anymore at this point :)

kennyweiss: I'm assuming that the order of the paths matters as well, right?

kennyweiss: Is there a evoke `$PATH` or `PYTHONPATH` -- i.e. variables with lists of directories where the order matters?

Maybe `spack_packages_path` would suffice if we make the connection in the docs?

(Somehow `spack_packages_path_list` is too specific in number of characters, but not specific enough about the order being important)

docs/sphinx/index.rst
```diff
@@ -154,26 +154,39 @@ copied to destination directory. This file specifies the compiler settings and p
...
+ ========================== ========================== ================================================ =======================================
+  package_name              ``--package-name``         Spack package name                               **None**
+  package_version           **None**                   Spack package version                            **None**
+  package_final_phase       ``--package-final-phase``  Controls after which phase Spack should stop     **None**
+  package_source_dir        ``--package-source-dir``   Controls the source directory Spack should use   **None**
+  spack_url                 **None**                   Url where to download Spack                      ``https://github.com/spack/spack.git``
+  spack_commit              **None**                   Spack commit to checkout                         **None**
+  spack_activate            **None**                   Spack packages to activate                       **None**
+  spack_packages_base_paths **None**                   List of Spack packages directories to be copied  ``packages``
```
kennyweiss, line 179: Please indicate that order matters (if it does).

docs/sphinx/index.rst
```diff
@@ -154,26 +154,39 @@ copied to destination directory. This file specifies the compiler settings and p
...
+  Setting                   Command line Option        Description                                      Default
+ ========================== ========================== ================================================ =======================================
+  package_name              ``--package-name``         Spack package name                               **None**
+  package_version           **None**                   Spack package version                            **None**
+  package_final_phase       ``--package-final-phase``  Controls after which phase Spack should stop     **None**
+  package_source_dir        ``--package-source-dir``   Controls the source directory Spack should use   **None**
+  spack_url                 **None**                   Url where to download Spack                      ``https://github.com/spack/spack.git``
+  spack_commit              **None**                   Spack commit to checkout                         **None**
+  spack_activate            **None**                   Spack packages to activate                       **None**
```
kennyweiss, line 178: Does `spack_configs_path` belong in this table?

kennyweiss (approved):
Thanks @white238  -- especially for the release-notes!

docs/sphinx/index.rst
white238: it now does and also it was in the documentation

docs/sphinx/index.rst
white238: Added... does this make sense to make this `spack_configs_paths` to be consistent with the packages variable?

white238: this would not be as simple as what I did... like what do you do if lives in both, do you take the first or last config?

white238: I'm not even sure this is helpful

white238:
Went back to the old variable name `spack_packages_path` and just put in some code to support a list/string and added a note that you can provide a list if you wish.  I think this is clearer because the most common use case is just a string.

adrienbernede (approved):
I think it makes sense to just call the variable "packages_path(s)" in the code, since we are in the spack specialized object. 👍🏻

uberenv.py
```diff
@@ -342,6 +358,18 @@ def __init__(self, opts, extra_opts):
 
         print("[spack spec: {}]".format(self.opts["spec"]))
 
+
+    def append_path_to_packages_paths(self, path, errorOnNonexistant=True):
+        path = pabs(path)
+        if not os.path.exists(path):
+            if errorOnNonexistant:
+                print("[ERROR: Given path in 'spack_packages_path' does not exist: {0}]".format(path))
```
white238, line 366: @kennyweiss I added a more clear error message here

uberenv.py
```diff
@@ -361,16 +389,30 @@ def setup_paths_and_dirs(self):
...
+            if type(_paths) is str:
+                # user gave a single string
+                self.append_path_to_packages_paths(_paths)
+            else:
+                # user gave a list of strings
+                for _path in _paths:
+                    self.append_path_to_packages_paths(_path)
+        else:
+            # default to packages living next to uberenv script if it exists
+            self.append_path_to_packages_paths(pjoin(self.uberenv_path,"packages"), errorOnNonexistant=False)
```
white238, line 405: I made this default path optional in case someone wanted to not provide their own packages

kennyweiss:
@adrienbernede  wrote:
> I think it makes sense to just call the variable "packages_path(s)" in the code, since we are in the spack specialized object. 👍🏻

I prefer the explicit `spack_` prefix since `vcpkg` is coming soon and we'll likely want/need to distinguish these.

white238:
Sorry for the churn but i think this is cleaner.

adrienbernede:
@kennyweiss I was commenting on a variable name _in the code_, local to SpackEnv object.

RELEASE-NOTES.md
```diff
@@ -0,0 +1,21 @@
...
+in this file.
+
+The format of this file is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
+
+The Uberenv project release numbers follow [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+### Added
+- Allow projects to force prefix on command line via new project option: `force_commandline_prefix`
```
kennyweiss, line 13: Could you please expand the "force prefix" part?

I.e. force the prefix for the library installation (or something a bit clearer).

white238: clarified.

uberenv.py
```diff
@@ -226,9 +226,13 @@ def parse_args():
...
 
 
+def pretty_print_dictionary(dictionary):
+    for key, value in dictionary.items():
+        print("  {0}: {1}".format(key, value))
+
 def uberenv_script_dir():
     # returns the directory of the uberenv.py script
-    return os.path.dirname(os.path.realpath(__file__))
+    return os.path.dirname(os.path.abspath(__file__))
```
white238, line 235: @kennyweiss this was causing problems when finding the config file in the parent directory when it undid the symlink.

white238:
merge?

uberenv.py
```diff
@@ -342,6 +358,18 @@ def __init__(self, opts, extra_opts):
...
+
+    def append_path_to_packages_paths(self, path, errorOnNonexistant=True):
+        path = pabs(path)
+        if not os.path.exists(path):
+            if errorOnNonexistant:
+                print("[ERROR: Given path in 'spack_packages_path' does not exist: {0}]".format(path))
+                sys.exit(1)
+            else:
+                return
+        self.packages_paths.append(pabs(path))
```
adrienbernede, line 370: path is already pabs(path).
```suggestion
        self.packages_paths.append(path)
```

PR #43, opened 2021-01-21
add error msg if given spack_configs_path is bad

cyrush:


cyrush:
thanks @kennyweiss 

(I did test this with a bad path, so I think it works as we expect)

Issue #45, opened 2021-01-21
spack spec is hard to read vs previous non verbose output


cyrush:

cyrush:
![Screen Shot 2021-01-21 at 2 14 46 PM](https://user-images.githubusercontent.com/1194526/105419169-22906100-5bf3-11eb-8388-cf19d8f162ff.png)

cyrush:
i have seen the light, having the detailed spec helps out

Issue #44, opened 2021-01-21
place created host config in --prefix dir


cyrush:
The used to be placed in the prefix dir, but now they placed in the dir where uberenv is run?
(I think we may have discussed this before)

I often use a few uberenv dirs -- I don't want the generated host configs to write over each other. 

Especially with the new strategy to stop at the hostconfig stage, b/c the install dir doesn't even exist -- if the host-config is written over, its gone. 

An option would be fine if folks are relying on this new behavior.

white238:
To my knowledge isn't handled at the uberenv level.  It is handled by the individual package's spack package.py.  Though we should have a unified solution to this.  This PR would help with this:

https://github.com/spack/spack/pull/19316/

cyrush:
with the old setup the host config was installed with the fake package for sure:

```
ls uberenv_libs/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.0-apple/uberenv-conduit-0.0.0-bgxrwhrdhehspvsulco5blmxeyru4ug5/
elder.llnl.gov-darwin-mojave-x86_64-clang@10.0.0-apple-conduit.cmake	lib
```

Then it also ended up at:
```
uberenv_libs/
elder.llnl.gov-darwin-mojave-x86_64-clang@10.0.0-apple-conduit.cmake
```

I expect the host-config to be generated in the staged dir -- but that is now transient, how does it end up getting written above the prefix by the spack package?
Sounds like I am doing something wrong. I'll look into it more.

cyrush:
For my case it was the fake package placing it:
`https://github.com/LLNL/conduit/blob/b3d1ba0feb41f4ed558a805a0023eef30b0485cf/scripts/uberenv/packages/uberenv-conduit/package.py#L79`

It's a bit scary to have the kind of logic in the proper spack package. I think uberenv should needs to help us here.

cyrush:
Good news: I think the logic in the spack package is fine.

It has to do with where we tell the `dev-build` to execute.

I'll think of a knob that makes sense.

cyrush:
```
        self.pkg_src_dir = os.path.join(self.uberenv_path,self.pkg_src_dir)
        if not os.path.isdir(self.pkg_src_dir):
            print("[ERROR: package_source_dir '{0}' does not exist]".format(self.pkg_src_dir))
            sys.exit(-1)
```

[...]

PR #46, opened 2021-01-23
2021-01 cyrush workarounds and tweaks

cyrush:
This branch includes:

1) A workaround to get back to point where the host config is generated in the prefix dir (#44)
2) Addition of a `-j` option to control build jobs (override what is presented in configs for more control)
3) Addition of a json file option to control (#45)

WIP.
I don't expect us to merge as is. Need a better solution for #44, we should discuss. 

Note: I am using this branch to allow me to move forward with updating uberenv for conduit.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -150,6 +150,11 @@ def parse_args():
                       dest="package_name",
                       default=None,
                       help="override the default package name")
+    # overrides uberenv_package_name
+    parser.add_option("--uberenv-package-name",
+                      dest="uberenv_package_name",
+                      default=None,
+                      help="override the default uberenv package name")
 
     # controls after which package phase spack should stop
     parser.add_option("--package-final-phase",
@@ -329,12 +334,15 @@ class UberEnv():
         else:
             print("[info: destination '{0}' already exists]".format(self.dest_dir))
 
-    def set_from_args_or_json(self,setting):
+    def set_from_args_or_json(self,setting, optional=True):
         try:
             setting_value = self.project_opts[setting]
         except (KeyError):
-            print("ERROR: '{0}' must at least be defined in project.json".format(setting))
-            raise
+            if optional:
+                print("ERROR: '{0}' must at least be defined in project.json".format(setting))
+                raise
+            else:
+                return None
         else:
             if self.opts[setting]:
                 setting_value = self.opts[setting]
@@ -502,8 +510,12 @@ class SpackEnv(UberEnv):
         UberEnv.__init__(self,opts,extra_opts)
 
         self.pkg_version = self.set_from_json("package_version")
-        self.pkg_final_phase = self.set_from_args_or_json("package_final_phase")
+        self.pkg_final_phase = self.set_from_args_or_json("package_final_phase",False)
         self.pkg_src_dir = self.set_from_args_or_json("package_source_dir")
+        # check if we are using uberenv fake package to setup dev env
+        self.uberenv_package_name = self.set_from_args_or_json("uberenv_package_name",False)
+        if not self.opts["install"] and self.uberenv_package_name:
+            self.pkg_name = self.uberenv_package_name
 
         self.packages_paths = []
 
@@ -601,18 +613,6 @@ class SpackEnv(UberEnv):
         if not os.path.isdir(self.pkg_src_dir):
             print("[ERROR: package_source_dir '{0}' does not exist]".format(self.pkg_src_dir))
             sys.exit(-1)
-
-        # # -------
-        # # dev-build is only used for the non install case (he hostconfig case)
-        # # can we set build dir relative to to dest?
-        # # -------
-        # self.pkg_src_dir = os.path.join(self.dest_dir,self.pkg_src_dir)
-        # if not os.path.isdir(self.pkg_src_dir):
-        #     os.mkdir(self.pkg_src_dir)
-        #     #print("[ERROR: package_source_dir '{0}' does not exist]".format(self.pkg_src_dir))
-        #     #sys.exit(-1)
-        
-
 
     def find_spack_pkg_path_from_hash(self, pkg_name, pkg_hash):
         res, out = sexe("spack/bin/spack find -p /{0}".format(pkg_hash), ret_output = True)
@@ -763,10 +763,15 @@ class SpackEnv(UberEnv):
 
     def show_info(self):
         # print concretized spec
+        # show short spec first
+        spec_cmd = "spack/bin/spack spec {1}{2}".format(self.pkg_name,self.opts["spec"])
+        res, out = sexe(spec_cmd, ret_output=True, echo=True)
+        print(out)
+        # now show detailed spec with install info
         # default case prints install status and 32 characters hash
         options="--install-status --very-long"
-        if not self.project_opts["spack_spec_print_options"] is None:
-            options = self.project_opts["spack_spec_print_options"]
+        #if not self.project_opts["spack_spec_print_options"] is None:
+        #    options = self.project_opts["spack_spec_print_options"]
         spec_cmd = "spack/bin/spack spec {0} {1}{2}".format(options,self.pkg_name,self.opts["spec"])
 
         res, out = sexe(spec_cmd, ret_output=True, echo=True)
@@ -799,12 +804,12 @@ class SpackEnv(UberEnv):
             if self.opts["ignore_ssl_errors"]:
                 install_cmd += "-k "
             if not self.opts["install"]:
-                # WE WANT HOST CONFIG TO LAND IN OUR UBERENV PREFIX DIR!
-                # THIS IS A HACK THAT HELPS ACCOMPLISH THAT
-                # The spack looks for this env var, if present
-                # uses it to select the destination. 
-                os.environ["UBERENV_HOSTCONFIG_DEST_DIR"] = self.dest_dir
-                install_cmd += "dev-build --quiet -d {0} ".format(self.pkg_src_dir)
+                if self.uberenv_package_name:
+                    # fake package path
+                    install_cmd += "-d install "
+                else:
+                    # dev build path
+                    install_cmd += "dev-build --quiet -d {0} ".format(self.pkg_src_dir)
                 if self.opts["build_jobs"]:
                     install_cmd += "-j {0}".format(self.opts["build_jobs"])
                 if self.pkg_final_phase:
@@ -846,6 +851,24 @@ class SpackEnv(UberEnv):
             res = sexe(activate_cmd, echo=True)
             if res != 0:
               return res
+        # if we used fake package path:
+        if not self.opts["install"] and not self.use_install and self.uberenv_package_name:
+            pkg_path = self.find_spack_pkg_path(self.pkg_name, self.opts["spec"])
+            if self.pkg_name != pkg_path["name"]:
+                print("[ERROR: Could not find install of {0}]".format(self.pkg_name))
+                return -1
+            else:
+                # Symlink host-config file
+                hc_glob = glob.glob(pjoin(pkg_path["path"],"*.cmake"))
+                if len(hc_glob) > 0:
+                    hc_path  = hc_glob[0]
+                    hc_fname = os.path.split(hc_path)[1]
+                    if os.path.islink(hc_fname):
+                        os.unlink(hc_fname)
+                    elif os.path.isfile(hc_fname):
+                        sexe("rm -f {0}".format(hc_fname))
+                    print("[symlinking host config file to {0}]".format(pjoin(self.dest_dir,hc_fname)))
+                    os.symlink(hc_path,hc_fname)
         # if user opt'd for an install, we want to symlink the final
         # install to an easy place:
         if self.opts["install"] or self.use_install:
--- a/uberenv.py
+++ b/uberenv.py
@@ -164,6 +164,12 @@ def parse_args():
                       default=pjoin(uberenv_script_dir(),"project.json"),
                       help="uberenv project settings json file")
 
+    # option to explicitly set the number of build jobs
+    parser.add_option("-j",
+                      dest="build_jobs",
+                      default=None,
+                      help="Explicitly set build jobs")
+
     # flag to use insecure curl + git
     parser.add_option("-k",
                       action="store_true",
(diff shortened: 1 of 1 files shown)


uberenv.py
```diff
@@ -814,9 +842,15 @@ def install(self):
                     print("[install complete!]")
         # otherwise we are in the "only dependencies" case and the host-config
         # file has to be copied from the do-be-deleted spack-build dir.
+        # -------
+        # NOTE: USING DEV BUILD, there is no `spack-build` dir
```
adrienbernede, line 846: @cyrush yes there is a `spack-build` dir, it is created as the source top dir.
The role of `dev-build` is to use the local version of the sources, that's why it is convenient: the host-config gets generated next to the local version of the source (in `src/spack-build`).
Then, uberenv moves the host-config file from `src/spack-build` to `src`, and removes `src/spack-build`.
If I understand correctly, that's where is becomes a problem for you: you may have several host-config files there.

cyrush: Correct. Overall, I want to avoid creating or deleting directories anywhere other than the prefix.

If things are limited to the prefix, I can avoid surprises.

cyrush: I though I checked to see if the `spack-build` was created for my case, and I didn't see evidence it happened (maybe b/c it was deleted before I could check) but I though I put in echos to see if ubrerenv saw it.  I can double check when I am back working on this.

uberenv.py
```diff
@@ -774,17 +792,25 @@ def show_info(self):
...
         # and build an host-config.cmake file
-
         if not self.use_install:
             install_cmd = "spack/bin/spack "
             if self.opts["ignore_ssl_errors"]:
                 install_cmd += "-k "
             if not self.opts["install"]:
-                install_cmd += "dev-build --quiet -d {0} ".format(self.pkg_src_dir)
+                if self.uberenv_package_name:
+                    # fake package path
```
white238, line 801: This isnt always a fake package.  This would also be helpful for installing an actual package that exists. Or am i missing something?

Like if i was in serac and wanted to build axom in their setup.

uberenv.py
```diff
@@ -150,6 +150,11 @@ def parse_args():
                       dest="package_name",
                       default=None,
                       help="override the default package name")
+    # overrides uberenv_package_name
+    parser.add_option("--uberenv-package-name",
```
white238, line 154: How is this different from --package-name that already exists?

uberenv.py
cyrush: if `uberenv_package_name`  exists, it doesn't use the dev build logic. that pretty much it

uberenv.py
cyrush: its presence triggers non dev build for the "non install" case (like ancient ubrerenv)

white238: --force-spack-install? 

then you would end up with `spack --package-name=fakepackage --force-spack-install` or just `spack --force-spack-install` which would end up with default package for the project.

My concern is how is a user supposed to know the difference between:

spack --package-name=foo
spack --uberenv-package-name=foo

without digging into the code.

cyrush: Better names are certainly welcome.

I think this will be set only in the json file for the post part. 

use case is to have uberenv work out of the box to build tpls when the `--install` option is not passed.

Having the option only be in the json file might be another option?

white238: Engage-Cyrus-Mode? =)

kennyweiss (commented):
@cyrush -- I think I get the general idea, but I am not sure I understand what the exact problem is and how this solves it.  

The logic here seems like it can get pretty hairy and difficult to maintain.

Is the problem that you want to control where the host-config gets generated?
If so, would it be simpler to have spack generate the host-config to its build/scratch directory
and allow uberenv to specify where we want to copy this file? (with a reasonable default)?

Or is there a different issue that I'm missing?

uberenv.py
```diff
@@ -774,17 +792,25 @@ def show_info(self):
...
+                else:
+                    # dev build path
+                    install_cmd += "dev-build --quiet -d {0} ".format(self.pkg_src_dir)
+                if self.opts["build_jobs"]:
+                    install_cmd += "-j {0}".format(self.opts["build_jobs"])
                 if self.pkg_final_phase:
                     install_cmd += "-u {0} ".format(self.pkg_final_phase)
             else:
                 install_cmd += "install "
+                if self.opts["build_jobs"]:
```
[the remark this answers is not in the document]
cyrush: good catch, i'll refactor

uberenv.py
```diff
@@ -323,12 +334,15 @@ def setup_paths_and_dirs(self):
...
             print("[info: destination '{0}' already exists]".format(self.dest_dir))
 
-    def set_from_args_or_json(self,setting):
+    def set_from_args_or_json(self,setting, optional=True):
         try:
             setting_value = self.project_opts[setting]
         except (KeyError):
-            print("ERROR: '{0}' must at least be defined in project.json".format(setting))
-            raise
+            if optional:
```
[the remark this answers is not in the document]
cyrush: my logic was wrong. (note this is still WIP)

I was surprised that many things are required in the json file, even though they may not be applicable. Like:

`pacakge_final_phase`

We have a check in the logic that applies it:

```
                if self.pkg_final_phase:
                    install_cmd += "-u {0} ".format(self.pkg_final_phase)

```

But you can't actually run uberenv unless it is there is a json file entry.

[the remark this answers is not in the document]
kennyweiss: Thanks @cyrush -- I agree that we should have sensible defaults, when applicable, to reduce the necessary entries in the `project.json` file

uberenv.py
cyrush: I removed this comment, not sure if its still an issue but it relates to a path I am not testing.

cyrush:
For context: I took several passes are using the dev build, including trying to inject fine grained control of the generated host config location, but I was not able to get parity with the original "uberenv fake" package approach. 

The goal for the fake package is support is to retain the workflow driving the setup of all the TPLs required for  development when uberenv is invoked with no args (relying on the project json file) and to also support deployment with install.

`uberenv_package_name` is the original json entry for uberenv, so I stuck with that. 


The "uberenv fake" package helps in the following ways:

1. Semantics of the fake package are just like any other spack package. Having a successfully installed fake package that we can query is helpful, we can look at things in the `.spack` dir to solve mysteries. We can actually uninstall the fake package on or way to re-gen the host config, etc. 

2.  Following best practices for Spack proper, the default variants don't line up with what we need out of the box for development.  With the fake package, we can make different choices (docs on, python on, etc) and support development out of the box. Consistent out of the box is very important -- we do development in ephemeral places (containers and scratch spaces) and have to bootstrap our dev env all the time.


3. The host config lives with the install in the prefix dir (and in the spack installed fake package dir). It is never written outside of the prefix or the spack stage area. This is really important for both development and deployment. I typically use between 2 and 4 uberenv prefix dirs  per project (replicated across several projects). My CMake work flow is basically "use this ubrerenv prefix dir" via using the host config inside. If I delete or move the my uberenv dirs I don't have to hunt down files elsewhere. Self contained greatly simplifies things.

Draw backs of the fake package approach:

1. Inheritance isn't really inheritance in spack, you can tangle with some meta programming magic or confusing variant tweaking to get a good state.

2. Some times it downloads things needlessly.


Since I have used the fake package approach for many years, the downsides are a known quantity. As I upgrade I am interested in vcpkg support and future spack env support, but dev build doesn't help me keep my current workflow.

I am not against folks using the dev build approach if that works for their dev workflow, but I would like to be to restore the fake package workflow, given it has been successful since the beginning of uberenv.

uberenv.py
cyrush: I am fine with removing this as an option from the command line.

uberenv.py
cyrush: I added `self.use_dev_build` as a better way of identifying the logic

cyrush:
also resolves:

https://github.com/LLNL/uberenv/issues/47

cyrush:
the non dev build (legacy) path is now only triggered by the`uberenv_package_name` json entry, there is no command line arg. Is that sufficiently hidden to move forward?

white238:
Is there something else `uberenv_package_name` does than just force install vs dev-build? If so why not have the option be a boolean that does just that? (will look over PR after I eat my food)

cyrush:
it specifies overridden package name  `uberenv-mypackage`

cyrush:
refactored input options to provide a `build-mode` arg and `build_mode` json entry. 

These options allow us to hand a specific build mode string down to the ubrerenv instances.  

refactored  spack instance logic to route this into `self.build_mode` with `dev-build` being the default.

Supported for spack are, `dev-build`, `uberenv-pkg`, and `install`. With this change folks can easily choose between dev build or uberenv pkg, via the json file, overridden by the command line arg.

uberenv.py
```diff
@@ -151,6 +151,13 @@ def parse_args():
                       default=None,
                       help="override the default package name")
 
+    # uberenv tpl build mode
+    parser.add_option("--build-mode",
+                      dest="build_mode",
+                      default=None,
+                      help="set mode used to build third party dependencies "
+                           "(spack options: 'dev-build' 'uberenv-pkg')")
```
kennyweiss, line 159: @cyrush -- a comment that you posted indicates that `install` is also an option.  Should it be added to this help line as well?

kennyweiss: This seems like a much cleaner way to expose this functionality!

kennyweiss: Looking through the code, it appears that `install` will be used if one does not set `--build-mode`,
so, perhaps the `help` should indicate that?

kennyweiss (approved):
Thanks @cyrush -- this is a lot cleaner than the previous version.

I think the logic for setting up the type of build is getting too complex and hard to follow, so we should consider if we can refactor that in the future.

It will be really helpful to have some automated CI tests on the uberenv user codes to support that refactor (as @adrienbernede is planning)

uberenv.py
```diff
@@ -323,23 +336,39 @@ def setup_paths_and_dirs(self):
...
+        setting_value = None
         try:
             setting_value = self.project_opts[setting]
         except (KeyError):
-            print("ERROR: '{0}' must at least be defined in project.json".format(setting))
-            raise
+            if not optional:
+                print("ERROR: '{0}' must at least be defined in project.json".format(setting))
+                raise
         else:
```
cyrush, line 353: i think this else needs to go

uberenv.py
cyrush: It's actually more complex than that :-(

The default `build_mode` should effectively be `dev-build`

The reason why it's complex has to do with the logic for command line args vs json overrides.

There is no one place for defaults, if you add a default in the command line arg def -- it will always override the json key (which isn't what we want).

adrienbernede:
> Thanks @cyrush -- this is a lot cleaner than the previous version.
> 
> I think the logic for setting up the type of build is getting too complex and hard to follow, so we should consider if we can refactor that in the future.
> 
> It will be really helpful to have some automated CI tests on the uberenv user codes to support that refactor (as @adrienbernede is planning)

There is a CI setup for Uberenv. It triggers a pipeline in both Serac and Conduit to check compatibility with the new Uberenv script. Watch me...

adrienbernede:
LGTM

adrienbernede:
And it will fail miserably for various reasons :(

cyrush:
@adrienbernede is there any good info to pull from the gitlab tests?
I'd like to get this in, but I want to make sure this doesn't break anything in serac.

adrienbernede:
The problem in Serac is that I just don't have access to their bank. Although I think they were OK with granting me access.

adrienbernede:
I'll see if I can get this to work.

adrienbernede:
I can confirm that this does not break Serac CI.

cyrush:
thanks @adrienbernede !

uberenv.py
```diff
@@ -151,6 +151,14 @@ def parse_args():
                       default=None,
                       help="override the default package name")
 
+    # uberenv tpl build mode
+    parser.add_option("--build-mode",
```
white238, line 155: spack-build-mode?

uberenv.py
```diff
@@ -494,13 +523,23 @@ class SpackEnv(UberEnv):
...
+        self.build_mode = self.set_from_args_or_json("build_mode",True)
+        # default spack build mode is dev-build
+        if self.build_mode is None:
+            self.build_mode = "dev-build"
+        # NOTE: install always overrides the build mode to "install"
+        if self.opts["install"]:
+            self.build_mode = "install"
+        # if we are using fake package mode, adjust the pkg name
+        if self.build_mode == "uberenv-pkg":
+            self.pkg_name =  "uberenv-" + self.pkg_name
```
white238, line 539: this seems to be project specific. Can package_name/package-name not be "uberenv-conduit" in your case?

uberenv.py
```diff
@@ -526,6 +565,13 @@ def __init__(self, opts, extra_opts):
 
         print("[spack spec: {0}]".format(self.opts["spec"]))
 
+    def print_spack_python_info(self):
+        spack_dir = self.dest_spack
+        cmd = pjoin(spack_dir,"bin","spack")
+        cmd += " python "
+        cmd += '-c "import sys; print(sys.executable);"'
```
white238, line 572: 👍

uberenv.py
```diff
@@ -848,21 +909,39 @@ def install(self):
                     os.symlink(pkg_path["path"],pabs(pkg_lnk_dir))
                     print("")
                     print("[install complete!]")
-        # otherwise we are in the "only dependencies" case and the host-config
-        # file has to be copied from the do-be-deleted spack-build dir.
         else:
```
white238, line 912: I am pretty sure this can be deindented
```
elif self.build_mode == "dev-build":
  ....
else: #original uberenv fake package case
 ...
```

white238 (approved):

uberenv.py
cyrush: I did have the exact name in there before, but that was the source of the confusing options name :-(

uberenv.py
cyrush: resolved this

uberenv.py
cyrush: i made this change

uberenv.py
white238: we talked .. we decided to keep this behaviour.

PR #48, opened 2021-03-02
Add ability to override vcpkg_ports, add error checking to spack_configs

white238:
Needed to move axom to a submodule.  Currently testing it.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -15,6 +15,7 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
   inside of your project's source repository.
 - Adds support for Windows builds using [Vcpkg].
 - Adds the `--triplet` command line argument for setting the Vcpkg build configuration.
+- Adds the `--vcpkg-ports-path` command line argument for setting the path to the vcpkg ports directory.
 
 ### Changed
 - Added ability to have multiple packages directories that will get copied into spack on top of
--- a/uberenv.py
+++ b/uberenv.py
@@ -380,9 +380,9 @@ class VcpkgEnv(UberEnv):
         else:
             # next to uberenv.py script (backwards compatibility)
             self.vcpkg_ports_path = pabs(pjoin(self.uberenv_path, "vcpkg_ports"))
-            _errmsg = "Could not find any directory for vcpkg ports. "
-                      "Use either command line option 'vcpkg-ports-path', "
-                      "config file option 'vcpkg_ports_path', or "
+            _errmsg = "Could not find any directory for vcpkg ports. " \
+                      "Use either command line option 'vcpkg-ports-path', " \
+                      "config file option 'vcpkg_ports_path', or " \
                       "defaulted directory 'vcpkg_ports' next to 'uberenv.py'"
 
         if not os.path.isdir(self.vcpkg_ports_path):
--- a/uberenv.py
+++ b/uberenv.py
@@ -373,11 +373,11 @@ class VcpkgEnv(UberEnv):
 
         # Find path to vcpkg ports
         _errmsg = ""
-        if self.opts["vcpkg_ports_path"] is not None:
+        if self.opts["vcpkg_ports_path"]:
             # Command line option case
             self.vcpkg_ports_path = pabs(self.opts["vcpkg_ports_path"])
             _errmsg = "Given path for command line option `vcpkg-ports-path` does not exist"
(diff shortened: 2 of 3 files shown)


uberenv.py
```diff
@@ -365,7 +365,29 @@ def setup_paths_and_dirs(self):
 
         UberEnv.setup_paths_and_dirs(self)
 
-        self.ports = pjoin(self.uberenv_path, "vcpkg_ports","*")
+        # Find path to vcpkg ports
```
kennyweiss, line 368: It seems like we should have a utility function to get the item from the command line with a fallback to uberenv_config

uberenv.py
```diff
@@ -517,8 +538,12 @@ def setup_paths_and_dirs(self):
 
         UberEnv.setup_paths_and_dirs(self)
 
-        # Spack yaml configs path (compilers.yaml, packages.yaml, etc.)
+        # Find Spack yaml configs path (compilers.yaml, packages.yaml, etc.)
+
+        # Next to uberenv.py (backwards compatility)
```
kennyweiss, line 543: 👍  for the comments

kennyweiss (approved):

uberenv.py
```diff
@@ -365,7 +365,29 @@ def setup_paths_and_dirs(self):
 
         UberEnv.setup_paths_and_dirs(self)
 
-        self.ports = pjoin(self.uberenv_path, "vcpkg_ports","*")
+        # Find path to vcpkg ports
+        _errmsg = ""
+        if not self.opts["vcpkg_ports_path"] is None:
```
adrienbernede, line 370: Just below is used the much more human-readable `if foo is not None`. I suggest we use the same syntax here.

uberenv.py
```diff
@@ -530,9 +555,14 @@ def setup_paths_and_dirs(self):
         # Test if the override option was used (--spack-config-dir)
         self.spack_config_dir = self.opts["spack_config_dir"]
         if self.spack_config_dir is None:
+            # If command line option is not used, search for platform under
+            # given directory
             uberenv_plat = self.detect_platform()
             if not uberenv_plat is None:
```
adrienbernede, line 561: Another candidate for the `is not None` syntax (out of scope, but I can see it... ;) ).

adrienbernede (approved):
Looks good to me. Only minor comments.

uberenv.py
white238: did this and all other instances in script.

uberenv.py
white238: I agree.  We should do a followup PR to do smaller cleanup things like this as well as add some CI checks to PRs.

white238:
@kennyweiss this seems to be passing the regular tests on axom. Could you test vcpkg since I cant?

RELEASE-NOTES.md
```diff
@@ -15,6 +15,7 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
   inside of your project's source repository.
 - Adds support for Windows builds using [Vcpkg].
 - Adds the `--triplet` command line argument for setting the Vcpkg build configuration.
+- Adds the `--vcpkg-ports-path` command line argument for setting the path to the vcpkg ports directory.
```
kennyweiss, line 18: I added a missing command line argument to support supplying a path to the vcpkg ports directory

uberenv.py
```diff
@@ -388,7 +416,7 @@ def clone_repo(self):
             clone_opts = ("-c http.sslVerify=false " 
                           if self.opts["ignore_ssl_errors"] else "")
 
-            clone_cmd =  "git {0} clone --single-branch --depth=1 -b {1} {2} vcpkg".format(clone_opts, vcpkg_branch,vcpkg_url)
+            clone_cmd =  "git {0} clone --single-branch -b {1} {2} vcpkg".format(clone_opts, vcpkg_branch,vcpkg_url)
```
kennyweiss, line 419: Note: `--depth=1` broke our ability to use a custom commit

kennyweiss (approved):

white238:
This was fully tested in Axom so should be fine to merge now.

uberenv.py
white238: Git cloning should also be merged into a function between the spack and vcpkg... next PR

uberenv.py
white238: logged #50 so we don't forget.

uberenv.py
white238: #49

PR #53, opened 2021-03-16
Remove wrong documentation

white238:


cyrush:
"%clang"" is still used as the default spec on macOS:

https://github.com/LLNL/uberenv/blob/fa5378549518560008ae345bd8a2e7ab01f8cb89/uberenv.py#L516

but yes -- the default config for darwin folder is no longer used.  

Side Note: Not sure why that was removed?

white238:
I'm not sure.  I just noticed when I was grep'ing in Serac for a path I missed moving to a submodule.

white238:
Fixed @cyrush suggestion

adrienbernede (approved):
LGTM

adrienbernede:
> "%clang"" is still used as the default spec on macOS:
> 
> https://github.com/LLNL/uberenv/blob/fa5378549518560008ae345bd8a2e7ab01f8cb89/uberenv.py#L516
> 
> but yes -- the default config for darwin folder is no longer used.
> 
> Side Note: Not sure why that was removed?

The Uberenv repo never had a `spack_configs` directory, the conduit version did. Or maybe I misunderstood the question?

white238:
> > "%clang"" is still used as the default spec on macOS:
> > https://github.com/LLNL/uberenv/blob/fa5378549518560008ae345bd8a2e7ab01f8cb89/uberenv.py#L516
> > 
> > but yes -- the default config for darwin folder is no longer used.
> > Side Note: Not sure why that was removed?
> 
> The Uberenv repo never had a `spack_configs` directory, the conduit version did. Or maybe I misunderstood the question?

I think uberenv (or at least one version did) defaulted the `spack_config_path` to `spack_configs/darwin` on OSX at some point.

cyrush:
hmm, looking closer -- I think the logic is still there?

https://github.com/LLNL/uberenv/blob/fa5378549518560008ae345bd8a2e7ab01f8cb89/uberenv.py#L568

cyrush:
here is the logic from ancient uberenv for comparison:

https://github.com/LLNL/conduit/blob/83ef572d7750e746ed9319b40cb8dc9126e13d8e/scripts/uberenv/uberenv.py#L170

adrienbernede:
Hmm, so yes, I misunderstood the question ;)

cyrush:
@adrienbernede correct -- sorry, I was referring to using the dir if it exists, not having the dir in the uberenv repo.

cyrush (approved):

PR #56, opened 2021-03-17
Deactivate CI: causing trouble while not usable yet

adrienbernede:
Gitlab CI has been disabled because:
- It needs to be fixed
- It causes quota issues (conduit CI is quite space consuming)
- It may cause too much traffic in projects pipeline views

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,18 +45,24 @@
 # Triggers Conduit pipeline asking for uberenv update
 # UPDATE_UBERENV variable, when set, will tell conduit to use the value as
 # a reference to the uberenv env revision to checkout on github.
-trigger-conduit:
-  variables:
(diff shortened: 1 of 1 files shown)


adrienbernede:
This is basically killing my workspace quota (number of inodes), because one run of conduit CI creates 500K files.

adrienbernede:
This is becoming an issue because the "LGTM" sign-off is no longer required for projects were only "trusted users" can contribute to with direct pushes.

cyrush (approved):

white238 (approved):

kennyweiss (approved):

adrienbernede:
I was hesitating, because I can make the serac workflow work. However, with the new Gitlab policy, this would be triggered on every commit to Uberenv, and this is just too much.

PR #58, opened 2021-03-24
fix bug with platform detection logic for spack case

### LINKED ISSUE #57: logic error in spack platform detection logic
right now, if you aren't running on LC (SYS_TYPE env set) or macos (darwin) you will get the error message:

`[ERROR: Given path for 'spack_configs_path' does not contain platform directories: ...]`

The logic here is wrong:

https://github.com/LLNL/uberenv/blob/92c4cb44b561530f8a03d8896945d6828fe0e07a/uberenv.py#L615

if platform is `None`, logic should simply skip trying to find the defaults for that platform.

(I hit this in a docker container)

cyrush:
resolves #57


white238 (approved):
So many paths.....

PR #60, opened 2021-04-12
bugfix for final symlink when complex specs are provided

cyrush:


uberenv.py
```diff
@@ -925,9 +926,11 @@ def install(self):
                     print("[removing project build directory {0}]".format(pjoin(build_dir)))
                     sexe("rm -rf {0}".format(build_dir))
             else: # original uberenv fake package case
-                pkg_path = self.find_spack_pkg_path(self.pkg_name, self.opts["spec"])
+                # use spec_hash to locate b/c other path won't work if
```
kennyweiss, line 929: Might make sense to try to refactor this a bit (in the future) since the logic's getting complex and there is at least some duplication in checks and error messages

kennyweiss (approved):
Thanks @cyrush !

uberenv.py
cyrush: i think i can do this here,  think the logic is now almost the same as the non dev build path.

cyrush: got this in, it's now the same logic as the install path so i removed the duplicated logic

PR #62, opened 2021-04-22
Add an option to select Spack's new concretizer

joshessman-llnl:
This approach uses the `json` config file to expose the clingo option, which triggers a modification of Spack's `config.yaml` in addition to an install of clingo through `pip`, packaged with https://github.com/joshessman-llnl/package-clingo-manylinux.

~I am in communication with the clingo dev(s) regarding an official PPC build that would allow us to switch over to the official package on PyPI.~ ✔️ This wheel-based approach was selected (instead of spack's bootstrap option) to avoid having to build clingo and its dependencies from source.

As an alternative interface, uberenv could read in the provided `config.yaml` (instead of the json option), see if it specifies the new concretizer, and install clingo accordingly.  Thoughts?

### DIFF
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -181,25 +181,25 @@ is found in the in the following order:
 
 Project settings are as follows:
 
- ========================== ========================== ================================================ =======================================
-  Setting                   Command line Option        Description                                      Default
- ========================== ========================== ================================================ =======================================
-  package_name              ``--package-name``         Spack package name                               **None**
-  package_version           **None**                   Spack package version                            **None**
-  package_final_phase       ``--package-final-phase``  Controls after which phase Spack should stop     **None**
-  package_source_dir        ``--package-source-dir``   Controls the source directory Spack should use   **None**
-  force_commandline_prefix  **None**                   Force user to specify `--prefix` on command line ``False``
-  spack_url                 **None**                   Download url for Spack                           ``https://github.com/spack/spack.git``
-  spack_commit              **None**                   Spack commit to checkout                         **None**
-  spack_activate            **None**                   Spack packages to activate                       **None**
-  spack_configs_path        **None**                   Directory with Spack configs to be copied        ``spack_configs``
-  spack_packages_path       **None**                   Directory with Spack packages to be copied       ``packages``
-  vcpkg_url                 **None**                   Download url for Vcpkg                           ``https://github.com/microsoft/vcpkg``
-  vcpkg_branch              **None**                   Vcpkg branch to checkout                         ``master``
-  vcpkg_commit              **None**                   Vcpkg commit to checkout                         **None**
-  vcpkg_ports_path          ``--vcpkg-ports-path``     Folder with vcpkg ports files                    **None**
-  spack_use_new_concretizer **None**                   Use Spack's ASP solver for concretization        ``False``
- ========================== ========================== ================================================ =======================================
+ ========================= ========================== ================================================ =======================================
+  Setting                  Command line Option        Description                                      Default
+ ========================= ========================== ================================================ =======================================
+  package_name             ``--package-name``         Spack package name                               **None**
+  package_version          **None**                   Spack package version                            **None**
+  package_final_phase      ``--package-final-phase``  Controls after which phase Spack should stop     **None**
+  package_source_dir       ``--package-source-dir``   Controls the source directory Spack should use   **None**
+  force_commandline_prefix **None**                   Force user to specify `--prefix` on command line ``False``
+  spack_url                **None**                   Download url for Spack                           ``https://github.com/spack/spack.git``
+  spack_commit             **None**                   Spack commit to checkout                         **None**
+  spack_activate           **None**                   Spack packages to activate                       **None**
+  spack_configs_path       **None**                   Directory with Spack configs to be copied        ``spack_configs``
+  spack_packages_path      **None**                   Directory with Spack packages to be copied       ``packages``
+  spack_concretizer        **None**                   Spack concretizer to use ``original, clingo``    ``original``
+  vcpkg_url                **None**                   Download url for Vcpkg                           ``https://github.com/microsoft/vcpkg``
+  vcpkg_branch             **None**                   Vcpkg branch to checkout                         ``master``
+  vcpkg_commit             **None**                   Vcpkg commit to checkout                         **None**
+  vcpkg_ports_path         ``--vcpkg-ports-path``     Folder with vcpkg ports files                    **None**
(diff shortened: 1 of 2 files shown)


uberenv.py
```diff
@@ -770,6 +776,17 @@ def patch(self):
                 print("[copying patched packages from {0}]".format(_src_glob))
                 sexe("cp -Rf {0} {1}".format(_src_glob, dest_spack_pkgs))
 
+        # Update spack's config.yaml if clingo was requested
+        if self.use_clingo:
+            import yaml
+            with open(pjoin(spack_etc_defaults_dir, "config.yaml"), 'r') as conf_file:
```
cyrush, line 782: Good that this import is deferred b/c pyyaml is not a part of all standard python installs.

uberenv.py
```diff
@@ -770,6 +776,17 @@ def patch(self):
...
+                conf_data = conf_file.read()
+            conf_yaml = yaml.load(conf_data)
+            conf_yaml['config']['concretizer'] = 'clingo'
+            # This will leave out all the comments, is this a problem?
+            with open(pjoin(spack_etc_defaults_dir, "config.yaml"), 'w') as conf_file:
```
cyrush, line 787: this is a copy that is placed in the uberenv spack dir right? I think its fine if comments are stripped. 

Q: could you check if the file already has `clingo` specified, and if so skip the injection?

uberenv.py
```diff
@@ -1035,6 +1052,39 @@ def use_spack_upstream(self):
...
+            py_interp = sys.executable
+            clingo_pkg = "clingo"
+            uninstall_cmd = "{0} -m pip uninstall -y {1}".format(py_interp, clingo_pkg)
+            # Uninstall it first in case the available version failed due to differing arch
+            # pip will still return 0 in the case of a "trivial" uninstall
+            res = sexe(uninstall_cmd, echo=True)
+            if res != 0:
+                print("[ERROR: clingo uninstall failed with returncode {0}]".format(res))
+                sys.exit(1)
+            install_cmd = "{0} -m pip install --user {1}".format(py_interp, clingo_pkg)
```
cyrush, line 1082: since this will be a user install, this breaks the mold of uberenv a bit. But not sure there is an easy alternative.  If the system install already has clingo this path is shortcuted so thats a good option for folks that want to avoid user install.

cyrush (approved):

uberenv.py
adrienbernede: I believe you could do
`spack config --scope defaults add config:concretizer:clingo`
and save yourself some trouble here.

adrienbernede (changes requested):
Interesting, I haven't used the new concretizer yet. Testing this will force me to ;)

uberenv.py
cyrush: using spack config command would avoid the pyyaml req for uberenv, so that's a really good adjustment

joshessman-llnl:
> 
> 
> Interesting, I haven't used the new concretizer yet. Testing this will force me to ;)

The biggest change is that you're required to provide a version for system ("external") packages - there's some discussion at https://github.com/LLNL/serac/pull/417#discussion_r616035913

cyrush (approved):

white238 (approved):

docs/sphinx/index.rst
```diff
@@ -181,24 +181,25 @@ is found in the in the following order:
...
+  spack_commit              **None**                   Spack commit to checkout                         **None**
+  spack_activate            **None**                   Spack packages to activate                       **None**
+  spack_configs_path        **None**                   Directory with Spack configs to be copied        ``spack_configs``
+  spack_packages_path       **None**                   Directory with Spack packages to be copied       ``packages``
+  vcpkg_url                 **None**                   Download url for Vcpkg                           ``https://github.com/microsoft/vcpkg``
+  vcpkg_branch              **None**                   Vcpkg branch to checkout                         ``master``
+  vcpkg_commit              **None**                   Vcpkg commit to checkout                         **None**
+  vcpkg_ports_path          ``--vcpkg-ports-path``     Folder with vcpkg ports files                    **None**
+  spack_use_new_concretizer **None**                   Use Spack's ASP solver for concretization        ``False``
```
kennyweiss, line 201: Future-proofing:  Would it be better to have:
```
spack_concretizer  { 'default', 'clingo', ... }
```

uberenv.py
kennyweiss: I tend to agree with @cyrush here.  

Directly calling `pip` on people's python seems different than the typical `uberenv` workflow. It might be preferable to error out here, with a message that tells people how to install `clingo` (and possibly provide some additional documentation about how to bootstrap their environment with `clingo`)

joshessman-llnl:
@adrienbernede's suggestion happens to not work - reported as https://github.com/spack/spack/issues/23259

kennyweiss (approved):
Thanks @joshessman-llnl 

Approving, but, as noted below, I have concerns about the workaround for getting `clingo`.  It seems like there might be a different/better way to use spack to bootstrap an environment with `clingo` ?

cyrush:
thats a bummer, hopefully they have a quick fix.

adrienbernede:
> @adrienbernede's suggestion happens to not work - reported as [spack/spack#23259](https://github.com/spack/spack/issues/23259)

What version of Spack are you testing with? Posted the command after testing it. I works on develop.
Maybe we have an issue with it in Uberenv context.

joshessman-llnl:
> 
> 
> > @adrienbernede's suggestion happens to not work - reported as [spack/spack#23259](https://github.com/spack/spack/issues/23259)
> 
> What version of Spack are you testing with? I posted the command after testing it. I works on develop.
> Maybe we have an issue with it in Uberenv context.

It looks like the patching Uberenv does with `config.py` is related, as it does seem to work without the patching

uberenv.py
kennyweiss: @joshessman-llnl -- I was thinking about this a bit more and wanted to clarify my position/review comments.

This is a tricky problem (i.e. requiring a dependency before you can run the dependency-getting tool). 

While I don't like the tool pip-installing things into the user's python, I'm ok with the proposed solution as a stop-gap until we can find something better. Especially since (a) this is an opt-in solution -- i.e. the user has to specifically ask for `clingo`, and (b) there are issues getting newer versions of spack to concretize without `clingo` in some workflows.


In terms of something better, would it be possible for uberenv to install `clingo` into a `virtualenv`, which it could then use as its python for spack?  My main concern is that installing into a user's system python can have unexpected/undesired side effects.

**tl;dr;** -- I'm ok with the current solution, and will welcome a better solution in the future

adrienbernede (approved):
Good to go!

PR #63, opened 2021-04-23
Feature/environments

adrienbernede:
Updates #18.

Missing:
- [ ] testing this, of course
- [x] force re-concretization. Either with `spack concretize -f`, or by removing the `spack.lock`. Most likely the latter.
- [ ] pass new mirror to the environment with spack command instead of config edit.
- [ ] copy environment in the uberenv_libs location to modify at will (see previous action).

The methodology here is to start with an empty environment: no spec.
Then uberenv adds the selected spec to the environment.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -733,7 +733,9 @@ class SpackEnv(UberEnv):
         # this is an opportunity to show spack python info post obtaining spack
         self.print_spack_python_info()
 
-
+# TODO: The hot copy may not be necessary any more, but we will deal with this
+#       after we implement the copy of environments and mirror setting.
+#       note: the feature to use here is: spack repo
         # hot-copy our packages into spack
         if len(self.packages_paths) > 0:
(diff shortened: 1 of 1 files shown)


cyrush:
The logic to setup the env is much simpler than the current patching which is nice. 

I hope we can keep the current logic and have an independent path for the env case. 

We can select the env path by the existence of a `spack.yaml` file, vs `compilers.yaml` etc, or better create a command line option `--spack-env` that takes the `spack.yaml` file directly.  Other than the env setup being different, I think the only other thing needed is to use  `-e .` during the spack install steps.  Also important is understanding how python packages are activated in the env case, I haven't look at that at all yet. 

Concretely: We should keep `patch` and add a case for the env logic. 

On the other TODOS / Questions: 

We should keep the print of the spack python info.  I have hit a few cases where folks were very confused by spack using a different python than uberenv was, that print helps us sort out cases like that. 

I also think we should keep the hot copy feature for the packages.

adrienbernede:
LGTM

PR #67, opened 2021-06-23
Change `spack clean` calls in Uberenv to preserve stage directories for debugging purposes

goxberry:
EDIT: Instead of making the changes described below, I followed
@white238's suggestion and changed the arguments to one of
the `spack clean` calls in Uberenv.

---------------

This commit adds a "--keep-stage" flag that disables the initial
`spack clean` and `spack clean --all` invocations when running
Uberenv. The name of this flag was chosen to evoke the same semantics
as the `--keep-stage` flag in `spack install`, `spack create`,
etc. However, this flag will *not* override the `--clean` flag of the
Uberenv CLI; I assume that if a user calls Uberenv with that flag,
they do intend to forcibly remove installations of previous packages
and their dependents.

The use case for this flag is the following: In one project I work on,
Uberenv is invoked repeatedly with different specs to install several
versions of third-party libraries. On this project, I would like to
keep the staged directories after the first invocation so that my
colleagues can drop into the source files of third-party libraries in
a debugger.

### DIFF
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -108,6 +108,7 @@ Uberenv has a few options that allow you to control how dependencies are built:
   ``--project-json``      File for project specific settings             See :ref:`project_configuration`
   ``--triplet``           (vcpkg) Target architecture and linkage        ``VCPKG_DEFAULT_TRIPLET`` environment variable,
                                                                          if present, ``x86-Windows`` otherwise
+  ``--keep-stage``        Keep stage dirs from previous Uberenv runs     **False**
  ======================= ============================================== ================================================
 
 The ``-k`` option exists for sites where SSL certificate interception undermines fetching
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -108,7 +108,6 @@ Uberenv has a few options that allow you to control how dependencies are built:
   ``--project-json``      File for project specific settings             See :ref:`project_configuration`
   ``--triplet``           (vcpkg) Target architecture and linkage        ``VCPKG_DEFAULT_TRIPLET`` environment variable,
                                                                          if present, ``x86-Windows`` otherwise
-  ``--keep-stage``        Keep stage dirs from previous Uberenv runs     **False**
  ======================= ============================================== ================================================
 
 The ``-k`` option exists for sites where SSL certificate interception undermines fetching
--- a/uberenv.py
+++ b/uberenv.py
@@ -230,14 +230,6 @@ def parse_args():
                       default=False,
                       help="Only download and setup Spack. No further Spack command will be run.")
 
-    # option to keep staged directories from previous Uberenv runs
-    parser.add_option("--keep-stage",
-                      action="store_true",
-                      dest="keep_stage",
-                      default=False,
-                      help="Keep stage directories from previous Uberenv runs;"
-                           "does not override --clean.")
-    
     ###############
     # parse args
     ###############
@@ -795,14 +787,10 @@ class SpackEnv(UberEnv):
 
 
     def clean_build(self):
-        if not self.opts["keep_stage"]:
-            # clean out any temporary spack build stages
-            cln_cmd = "spack/bin/spack clean "
-            res = sexe(cln_cmd, echo=True)
-
-            # clean out any spack cached stuff
-            cln_cmd = "spack/bin/spack clean --all"
-            res = sexe(cln_cmd, echo=True)
+        # clean out any spack cached stuff (except build stages, downloads, &
+        # spack's bootstrapping software)
+        cln_cmd = "spack/bin/spack clean --misc-cache --failures --python-cache"
+        res = sexe(cln_cmd, echo=True)
 
         # check if we need to force uninstall of selected packages
         if self.opts["spack_clean"]:
--- a/uberenv.py
+++ b/uberenv.py
@@ -796,13 +796,13 @@ class SpackEnv(UberEnv):
 
     def clean_build(self):
         if not self.opts["keep_stage"]:
-           # clean out any temporary spack build stages
-           cln_cmd = "spack/bin/spack clean "
-           res = sexe(cln_cmd, echo=True)
+            # clean out any temporary spack build stages
+            cln_cmd = "spack/bin/spack clean "
+            res = sexe(cln_cmd, echo=True)
 
-           # clean out any spack cached stuff
-           cln_cmd = "spack/bin/spack clean --all"
-           res = sexe(cln_cmd, echo=True)
+            # clean out any spack cached stuff
+            cln_cmd = "spack/bin/spack clean --all"
+            res = sexe(cln_cmd, echo=True)
 
         # check if we need to force uninstall of selected packages
         if self.opts["spack_clean"]:
--- a/uberenv.py
+++ b/uberenv.py
@@ -230,6 +230,7 @@ def parse_args():
                       default=False,
                       help="Only download and setup Spack. No further Spack command will be run.")
 
+
     ###############
     # parse args
     ###############
--- a/uberenv.py
+++ b/uberenv.py
@@ -230,7 +230,14 @@ def parse_args():
                       default=False,
                       help="Only download and setup Spack. No further Spack command will be run.")
 
-
+    # option to keep staged directories from previous Uberenv runs
+    parser.add_option("--keep-stage",
+                      action="store_true",
+                      dest="keep_stage",
+                      default=False,
+                      help="Keep stage directories from previous Uberenv runs;"
+                           "does not override --clean.")
+    
     ###############
     # parse args
     ###############
@@ -788,13 +795,14 @@ class SpackEnv(UberEnv):
 
 
     def clean_build(self):
-        # clean out any temporary spack build stages
-        cln_cmd = "spack/bin/spack clean "
-        res = sexe(cln_cmd, echo=True)
+        if not self.opts["keep_stage"]:
+           # clean out any temporary spack build stages
+           cln_cmd = "spack/bin/spack clean "
+           res = sexe(cln_cmd, echo=True)
 
-        # clean out any spack cached stuff
-        cln_cmd = "spack/bin/spack clean --all"
-        res = sexe(cln_cmd, echo=True)
+           # clean out any spack cached stuff
+           cln_cmd = "spack/bin/spack clean --all"
+           res = sexe(cln_cmd, echo=True)
 
         # check if we need to force uninstall of selected packages
         if self.opts["spack_clean"]:


white238:
It was my understanding that we were utilizing the `spack clean` command to remove the temporary and cache files but the comment clearly states otherwise.  I would prefer to leave the staging directories by default.

After reading the documentation here:

https://spack.readthedocs.io/en/latest/packaging_guide.html?highlight=clean#spack-clean

I would prefer to just lock down the command to something more like `spack clean --misc-cache --failures --python-cache` and then not have another option.

Am I missing a side effect this may cause, @cyrush ?

goxberry:
@white238 Yes, that's a better idea.

goxberry:
@white238 Done.

kennyweiss (approved):
Thanks @goxberry  and @white238

goxberry:
@kennyweiss @white238 @cyrush Is there anything I can do to help this PR land in Uberenv? One of the projects I work on is waiting for this PR to land. I can reconfigure our uberenv git submodule instance to point to my fork, but we have several junior developers working on that project, so if I can avoid that reconfiguration, it would save me the trouble of having to debug their setups with them.

cyrush:
My perspective: aggressive cleaning (`spack clean --all`) help avoid surprises when you rerun uberenv. 

I am fine with an option to keep things around -- but I think it should be an option, I don't think we should remove the clean --all.

cyrush:
Ok, I am a little confused: Don't we just need `--keep-stage` on the ``install`` command?

That seems totally orthogonal from the clean settings?

white238:
To my understanding, `spack clean --all` removes previous stage directories.

https://spack.readthedocs.io/en/latest/command_index.html#spack-clean

This is the equivalent to something like `spack clean --all --keep-stage` but that doesnt exist.

cyrush:
Ok, so if you run uberenv -- everything installs fine, then you run uberenv again it will clean up the stages?

I think you still need --keep-stage for successful builds?  We aren't calling clean after install.

white238:
> 
> 
> Ok, so if you run uberenv -- everything installs fine, then you run uberenv again it will clean up the stages?
> 
> I think you still need --keep-stage for successful builds? We aren't calling clean after install.

correct

cyrush:
Ok, lets aim for the clean change -- but I still think we need `--keep-stage`?

I think Spack's behavior has changed over time (added new things that are cached, etc)

It would help to have a few test runs shared so we can verify everything works like we expect.

goxberry:
> I think you still need `--keep-stage` for successful builds? We aren't calling clean after install.

We're using `dev-build` mode, and `spack dev-build` does not have a `--keep-stage` option. You're right that `--keep-stage` would need to be added to `spack install` invocations.

goxberry:
> It would help to have a few test runs shared so we can verify everything works like we expect.

Is there a test suite somewhere?

cyrush:
Let's make the change.


Spack's clean behavior has changed over time.
`spack clean` did not always equate to `spack clean --all`, however that is now the case. 

This was throwing me off.  Clean `--all` wasn't even removing the stages in the older version I was testing. 
I also think that dev-build might have different rules. 

To move forward, I was just looking for an example to make sure we have a handle on the current spack behavior. 
We *do* need `--keep-stage` for any non dev build cases. I added a ticket for this: #68.

Here is an example of the difference with when installing.
(with `--keep-stage`)

```
[harrison37@elder spack (develop)]$ bin/spack clean --all 
==> Removing all temporary build stages
==> Removing cached downloads
==> Removing install failure marks
==> Removing cached information on repositories
==> Removing python cache files
==> Removing software in "/Users/harrison37/.spack/bootstrap/store"
[harrison37@elder spack (develop)]$ bin/spack install --keep-stage libxml2 
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/libiconv-1.16-2py5rvx5z25udmw3g5463rimshfkqqxa
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/pkgconf-1.7.4-riy5ze5ht5ch2ewfjpkgocdmr7iplxc5
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/xz-5.2.5-r6r6ntbbrxa4vwfifcrbpv3cbc6tfc7y
==> Installing zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw
==> No binary for zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw found: installing from source
==> Warning: microarchitecture specific optimizations are not supported yet on mixed compiler toolchains [check apple-clang@11.0.0 for further details]
==> Fetching https://mirror.spack.io/_source-cache/archive/c3/c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1.tar.gz
==> No patches needed for zlib
==> zlib: Executing phase: 'install'
==> zlib: Successfully installed zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw
  Fetch: 3.21s.  Build: 4.19s.  Total: 7.39s.
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw
==> Installing libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g
==> No binary for libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g found: installing from source
==> Warning: microarchitecture specific optimizations are not supported yet on mixed compiler toolchains [check apple-clang@11.0.0 for further details]
==> Fetching https://mirror.spack.io/_source-cache/archive/aa/aafee193ffb8fe0c82d4afef6ef91972cbaf5feea100edc2f262750611b4be1f.tar.gz
==> Fetching https://mirror.spack.io/_source-cache/archive/96/96151685cec997e1f9f3387e3626d61e6284d4d6e66e0e440c209286c03e9cc7.tar.gz
==> Moving resource stage
	source : /var/folders/wj/jjhjt_0d3kg1vmk56d5bq_38001kr8/T/harrison37/spack-stage/resource-xmlts-mrq6i4jb2t5zoozreokxqoeixhrc5l4g/spack-src/
	destination : /var/folders/wj/jjhjt_0d3kg1vmk56d5bq_38001kr8/T/harrison37/spack-stage/spack-stage-libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g/spack-src/xmlconf
==> Ran patch() for libxml2
==> libxml2: Executing phase: 'autoreconf'
==> libxml2: Executing phase: 'configure'
==> libxml2: Executing phase: 'build'
==> libxml2: Executing phase: 'install'
==> libxml2: Successfully installed libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g
  Fetch: 8.25s.  Build: 57.06s.  Total: 1m 5.32s.
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g
[harrison37@elder spack (develop)]$ bin/spack location zlib
/var/folders/wj/jjhjt_0d3kg1vmk56d5bq_38001kr8/T/harrison37/spack-stage/spack-stage-zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw/spack-src
[harrison37@elder spack (develop)]$ bin/spack location libxml2
/var/folders/wj/jjhjt_0d3kg1vmk56d5bq_38001kr8/T/harrison37/spack-stage/spack-stage-libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g/spack-src
[harrison37@elder spack (develop)]$ bin/spack clean
==> Removing all temporary build stages
[harrison37@elder spack (develop)]$ bin/spack location zlib
==> Error: Source directory does not exist yet. Run this to create it:
  spack stage zlib
[harrison37@elder spack (develop)]$ bin/spack location libxml2
==> Error: Source directory does not exist yet. Run this to create it:
  spack stage libxml2
```


(without `--keep-stage`)
```
[harrison37@elder spack (develop)]$ bin/spack uninstall --dependents zlib 
==> The following packages will be uninstalled:

    -- darwin-mojave-skylake / apple-clang@11.0.0 -------------------
    mrq6i4j libxml2@2.9.10  vehd5ds zlib@1.2.11

==> Do you want to proceed? [y/N] y
==> Successfully uninstalled libxml2@2.9.10%apple-clang@11.0.0~python arch=darwin-mojave-skylake/mrq6i4j
==> Successfully uninstalled zlib@1.2.11%apple-clang@11.0.0+optimize+pic+shared arch=darwin-mojave-skylake/vehd5ds
[harrison37@elder spack (develop)]$ bin/spack install libxml2 
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/libiconv-1.16-2py5rvx5z25udmw3g5463rimshfkqqxa
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/pkgconf-1.7.4-riy5ze5ht5ch2ewfjpkgocdmr7iplxc5
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/xz-5.2.5-r6r6ntbbrxa4vwfifcrbpv3cbc6tfc7y
==> Installing zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw
==> No binary for zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw found: installing from source
==> Warning: microarchitecture specific optimizations are not supported yet on mixed compiler toolchains [check apple-clang@11.0.0 for further details]
==> Using cached archive: /Users/harrison37/Work/spack/var/spack/cache/_source-cache/archive/c3/c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1.tar.gz
==> No patches needed for zlib
==> zlib: Executing phase: 'install'
==> zlib: Successfully installed zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw
  Fetch: 0.00s.  Build: 4.64s.  Total: 4.65s.
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/zlib-1.2.11-vehd5dsoagzcexij7zkjr7tb7que36yw
==> Installing libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g
==> No binary for libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g found: installing from source
==> Warning: microarchitecture specific optimizations are not supported yet on mixed compiler toolchains [check apple-clang@11.0.0 for further details]
==> Using cached archive: /Users/harrison37/Work/spack/var/spack/cache/_source-cache/archive/aa/aafee193ffb8fe0c82d4afef6ef91972cbaf5feea100edc2f262750611b4be1f.tar.gz
==> Using cached archive: /Users/harrison37/Work/spack/var/spack/cache/_source-cache/archive/96/96151685cec997e1f9f3387e3626d61e6284d4d6e66e0e440c209286c03e9cc7.tar.gz
==> Moving resource stage
	source : /var/folders/wj/jjhjt_0d3kg1vmk56d5bq_38001kr8/T/harrison37/spack-stage/resource-xmlts-mrq6i4jb2t5zoozreokxqoeixhrc5l4g/spack-src/
	destination : /var/folders/wj/jjhjt_0d3kg1vmk56d5bq_38001kr8/T/harrison37/spack-stage/spack-stage-libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g/spack-src/xmlconf
==> Ran patch() for libxml2
==> libxml2: Executing phase: 'autoreconf'
==> libxml2: Executing phase: 'configure'
==> libxml2: Executing phase: 'build'
==> libxml2: Executing phase: 'install'
==> libxml2: Successfully installed libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g
  Fetch: 0.03s.  Build: 56.67s.  Total: 56.70s.
[+] /Users/harrison37/Work/spack/opt/spack/darwin-mojave-skylake/apple-clang-11.0.0/libxml2-2.9.10-mrq6i4jb2t5zoozreokxqoeixhrc5l4g
[harrison37@elder spack (develop)]$ bin/spack location zlib 
==> Error: Source directory does not exist yet. Run this to create it:
  spack stage zlib
```

cyrush (approved):

adrienbernede:
@becker33
I have one concern here (as @cyrush said, we don’t want any surprises).

(In advance sorry for the confusion between stage, cache, and downloads).

Here is a simplified scenario:
I fetch (or stage) `foo@develop`, then try to build it later, while develop branch changed in the upstream repo.

Will Spack build the newer develop, or the older (still staged / cached) one?

In particular, I think we want to always point to the actual upstream develop. And cleaning the downloads and stage seemed to guarantee that. What if we don't clean the downloads and / or stage?

goxberry:
@adrienbernede The trouble with the suggestion

> In particular, I think we want to always point to the actual upstream develop. And cleaning the downloads and stage seemed to guarantee that. What if we don't clean the downloads and / or stage?

is that in the use case discussed at the top of this thread, cleaning the stage directories at the start of every Uberenv invocation means that stage directories -- and thus debugging information -- does not persist. It is the "always" part of that suggestion that is problematic, and causes problems when using uberenv in scripting workflows to support development of projects that need multiple TPL configurations with debugging information. Unless there were some way to override that hypothetical default with a flag -- which I will note was the original implementation in this pull request -- the project I work on would not be able to use uberenv to build multiple TPLs with debugging information, which has been a significant pain point.

cyrush:
@goxberry 

(maybe this is moot b/c we merged the PR)

The use case is browsing the source code with debugging info?  Is there a reason why debugging symbols aren't sufficient for your workflow? Are the libs you can't build with debugging symbols?

adrienbernede:
@goxberry I completely agree with you: my 'always' excluded your use case.

My comment means: I prefer the original implementation. BUT, I wanted to double check with Greg that my concerns were founded. Also I _think_ in the near future Spack will better handle git and address in particular the scenario I was describing.

goxberry:
> (maybe this is moot b/c we merged the PR)

> The use case is browsing the source code with debugging info? Is there a reason why debugging symbols aren't sufficient for your workflow? Are the libs you can't build with debugging symbols?

Yes, the use case is browsing the source code within a debugger. I am not an expert in debuggers, and my understanding is that debug symbols should be sufficient, but here is what I see when I try to debug an executable that is built in debug mode, uses MFEM built with debugging symbols, and place a breakpoint in `mfem::DataCollection::Save`, which is defined in the `fem/datacollection.cpp` file in MFEM (and thus is declared but not defined in a header):

Screen Shot 2021-07-19 at 12 06 45

I know that `libmfem.a` has debug symbols in it because I ran `readelf --debug-dump=info` on it:

Screen Shot 2021-07-19 at 12 12 01

so it appears that the source code is required for browsing in a debugger, at least for GDB, and possibly other debuggers.

There aren't any libraries that project currently uses that cannot be built with debug symbols (that we care about, anyway).

PR #71, opened 2021-11-16
--resuse and --upstream w/ --setup-only.

mdavis36:
# Summary

This PR adds --reuse for use with spack v0.17+. It also enables upstream compatibility when using --setup-only.

#### --reuse
 - Spack provides the `--reuse` option to `spec` and `install` commands for more relaxed tpl/dependency matching. A new uberenv argument has been added to enable this. 

#### Upstream
 - --upstream now works when using --setup-only mode.


uberenv.py
```diff
@@ -736,7 +742,8 @@ def patch(self):
         self.print_spack_python_info()
 
         # force spack to use only "defaults" config scope
-        self.disable_spack_config_scopes(spack_dir)
+        os.environ["SPACK_DISABLE_LOCAL_CONFIG"] = "1"
```
adrienbernede, line 745: For the records, I think this is not what we want:
With this, the config scope is disabled only in the script context. That means the uberenv spack instance can get "polluted" by an out of script use.

mdavis36: This PR is fairly out of date with develop, it looks like uberenv is currently implementing this technique in develop for newer versions of spack.

uberenv.py
```diff
@@ -1202,6 +1199,9 @@ def main():
 
         return 0
 
+    # Show the spec for what will be built
+    env.show_info()
+
```
adrienbernede, line 1204: Looking at this, I wonder:
If the result of the concretization depends on what is already installed (--reuse), shouldn’t the concretization be done after the mirrors and upstream are set ?
In other words, I worry that reordering things that way could show a different concretization than what will actually be built.
What do you think?

mdavis36: This is a change back to what is already in uberenv. This call to `show_info()` is not modified by this PR. I agree however that `show_info()` might make more sense to be called after upstreams / mirrors are set.

mdavis36:
@adrienbernede Is this good to go or would you like this PR to change the `show_info()` logic? If so, what would be your preferred solution?

uberenv.py
```diff
@@ -812,7 +818,7 @@ def clean_build(self):
 
         # check if we need to force uninstall of selected packages
         if self.opts["spack_clean"]:
-            if self.project_opts.has_key("spack_clean_packages"):
+            if "spack_clean_packages" in self.project_opts:
```
white238, line 821: 👍

uberenv.py
```diff
@@ -1179,6 +1193,10 @@ def main():
 
     # Allow to end uberenv after spack is ready
     if opts["setup_only"]:
+
+        if opts["upstream"] is not None:
```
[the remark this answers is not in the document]
white238: This, as well as the one below, will fail if running vcpkg instead of Spack. Can you either add these two guards against windows or add an error check in the function that parse's arguments here

https://github.com/LLNL/uberenv/blob/0e95ce0130db47aa015e8ca466cf038b2d76adca/uberenv.py#L250

kennyweiss (approved):
Thanks @mdavis36 -- this looks like a nice addition and reasonably harmless when not used

What happens if you add `reuse` with an older spack?
Does it ignore the extra command? or fail with an error message about an unknown option? 

Please update the `RELEASE_NOTES` before merging.

mdavis36:
@kennyweiss Done! I believe using reuse with an older spack will give you an error message from spack about the unrecognized option.

cyrush (approved):

white238:
Thanks @mdavis36 !

adrienbernede:
Thank you @mdavis36, no worries regarding `show_info`.

adrienbernede:
@kennyweiss @mdavis36 confirmed behavior with older spack:
```
[exe: spack/bin/spack dev-build --reuse --quiet -d /usr/WS1/radiuss/tests/RAJA -u hostconfig raja@develop%gcc]
==> Error: unrecognized arguments: --reuse
[ERROR: failure of spack install/dev-build]
```

PR #74, opened 2021-12-22
new disable scopes support and basic github actions ci

cyrush:


uberenv.py
```diff
@@ -719,11 +720,25 @@ def disable_spack_config_scopes(self,spack_dir):
...
+        #
+        # For newer versions of spack, we can use the SPACK_DISABLE_LOCAL_CONFIG
+        # env var plumbing. We patch it to True to make a permanent change.
+        #
+        # Note: This path does not disable the 'site' config, but disabling 'user' config
+        # is our primary goal.
+        #
+        spack_disable_env_stmt = 'disable_local_config = "SPACK_DISABLE_LOCAL_CONFIG" in os.environ'
+        spack_disable_env_stmt_perm = "disable_local_config = True"
+        if cfg_script.count(spack_disable_env_stmt) > 0:
```
white238, line 732: 👍

cyrush:
I added very basic ci testing as well. We can expand as needed.

cyrush:
resolves #74

white238 (approved):

white238:
After this gets merged into `master`, we can force that the checks need to pass before being able to merge PRs.

.ci/test-project/uberenv_config.json
```diff
@@ -0,0 +1,10 @@
+{
+"package_name" : "magictestlib",
+"package_version" : "1.0.0",
+"package_source_dir" : ".",
+"spack_build_mode" : "uberenv-pkg",
+"spack_configs_path": "spack_configs",
+"spack_packages_path": "packages",
+"spack_url": "https://github.com/spack/spack",
+"spack_branch": "develop"
```
kennyweiss, line 9: Do we want to tie this to `spack@develop` or  lock it down to a specific commit/tag?

.github/workflows/linux.yml
```diff
@@ -0,0 +1,36 @@
...
+                                curl \
+                                wget \
+                                tar \
+                                unzip \
+                                build-essential
+    - uses: actions/checkout@v2
+    - name: Run Uberenv
+      run: |
+          cd  .ci/test-project
+          python ../../uberenv.py  --project-json=uberenv_config.json
```
kennyweiss, line 32: Does this fail if uberenv fails? or does it need some logic to fail the CI?

.ci/test-project/packages/uberenv-magictestlib/package.py
```diff
@@ -0,0 +1,48 @@
+##############################################################################
```
kennyweiss, line 1: Is this one for compatibility with the older workflows?

.ci/test-project/packages/magictestlib/package.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    if vtype is None:
+        if value == "ON" or value == "OFF":
+            vtype = "BOOL"
+        else:
+            vtype = "PATH"
+    return 'set({0} "{1}" CACHE {2} "")\n\n'.format(name, value, vtype)
+
+
+class Magictestlib(Package):
```
kennyweiss, line 30: Should this extend from `CachedCMakePackage` ? 
Or were you going for simplicity?

kennyweiss (approved):
Thanks @cyrush

.ci/test-project/packages/magictestlib/package.py
cyrush: I don't want to build cmake. Just want to test the basic mechanics.

.ci/test-project/packages/uberenv-magictestlib/package.py
cyrush: this test is based on how i use uberenv across projects, but we can also add things that test the dev build but I am not the expert on that way of using.

.github/workflows/linux.yml
cyrush: the ci fails uberenv fails (or if the cat command fails)

.ci/test-project/uberenv_config.json
cyrush: I say we pin *if* we see problems.  The hope is that this is so basic that spack changes shouldn't be able to undermine it.  That might not be the case when we add things like envs, so we can reconsider later.

PR #75, opened 2021-12-23
WIP spack envs support

cyrush:
Proof of concept -- targeted surgery to get a host config build using a spack env.

I am not totally sold on benefits from this support - we should discuss more.

There could be a win when we have specs that we want to customize via the env, or extra support we always want on for development (like +docs) -- however more and more version constraints will have to be in spack proper b/c their CI now depends on them (which is a good thing!)


uberenv.py
```diff
@@ -542,6 +556,7 @@ def __init__(self, opts, extra_opts):
         self.packages_paths = []
         self.spec_hash = ""
         self.use_install = False
+        self.use_env = False
```
cyrush, line 559: - [ ] remove this

uberenv.py
```diff
@@ -782,9 +840,9 @@ def patch(self):
 
             if os.path.isfile(packages_yaml):
                 sexe("cp {0} {1}/".format(packages_yaml, spack_etc_defaults_dir ), echo=True)
-        else:
+        elif env_file is None:
```
cyrush, line 843: - [ ] use helper method to det case

uberenv.py
```diff
@@ -874,13 +959,13 @@ def install(self):
                 return -1
             if self.opts["build_jobs"]:
                 install_cmd += "-j {0} ".format(self.opts["build_jobs"])
-            # for all cases we use the pkg name and spec
-            install_cmd += self.pkg_name + self.opts["spec"]
+            # for almost all cases we use the pkg name and spec
+            if self.spack_env is None:
```
cyrush, line 963: - [ ] use helper method to detect case

white238:
information from @becker33:

Yeah, if you're willing to put a matrix in your environment and specify all the compilers you want, that's easy
 

specs:
  - matrix:
    - [hdf5, mypkg]
    - ['%clang@10.0.0', '%intel@19.0.4']

But if you want not to specify the compiler in the environment, and only specify it at runtime, then you need something like
 

specs: [hdf5, mypkg]

and

spack -c packages:all:require:%clang@10.0.0 -e my_env concretize -f; spack -e my_env install
spack -c packages:all:require:%intel@19.0.4 -e my_env concretize -f; spack -e my_env install

which isn't 100% satisfying, but should work until we can implement something better.

cyrush:
Q: With a fused env, will all compilers and compiler specifics end up looking like?

becker33:
From the same conversation:

> Oh the other thing you could do for uberenv is something like:
 
```
spack env create temp_intel /path/to/spack.yaml
spack env activate temp_intel
spack config add packages:all:require:%intel@19.0.4
spack install
```

> Where the compiler config is never added to the environment proper, but you spin up temporary environments that get the last bit of configuration thrown in.

becker33:
@cyrush I'm not sure I follow the question -- the specs will be differentiated as usual by their hashes and the attributes that go into those hashes.

cyrush:
It would be helpful If I typed the words that I was thinking.

Just wondering what an everything in one file env would end up looking like.

Trying to think of a way to combine package specs, external compiler and package info sent me into a tailspin in the past. 

Env per-system per-compiler seems not great, but might still be were we need to start.

becker33:
For all-in-one, we have the matrix syntax. That's what we use for TCE.

```
specs:
- matrix:
  - [package1, package2]
  - [^mpi1, ^mpi2]
  - ['%compiler1', '%compiler2']
```

that can also be refactored with named lists and definitions, but that's a starting point.

white238:
It is not as simple as a combination of lists.  Here is an example of part of the specs that Axom needs for one SYS_TYPE:

```
 [ "clang@9.0.0+devtools+mfem+c2c",
      "clang@10.0.0+devtools+mfem+c2c+scr ^scr@3.0.1~shared~tests~examples",
      "gcc@8.3.1+devtools+mfem+c2c+scr ^scr@3.0.1~shared~tests~examples",
      "gcc@8.1.0~fortran+devtools~mfem+c2c+scr ^scr@3.0.1~shared~tests~examples",
      "intel@19.0.4~openmp+devtools+mfem+c2c" ],
```

becker33:
There's no restriction on the level of complexity at any level of the matrix, could be something like...

```
specs:
- matrix:
  - [axom~cuda+devtools+c2c]
  - [ "%clang@9.0.0+mfem",
      "%clang@10.0.0+mfem+scr ^scr@3.0.1~shared~tests~examples",
      "%gcc@8.3.1+mfem+scr ^scr@3.0.1~shared~tests~examples",
      "%gcc@8.1.0~fortran~mfem+scr ^scr@3.0.1~shared~tests~examples",
      "%intel@19.0.4~openmp+mfem" ]
  - [arch=linux-rhel7-x86_64_v2]
- matrix:
  - [axom+cuda+devtools+c2c]
  - [ "%clang@8.0.1+mfem+cuda cuda_arch=70",
      "%gcc@8.3.1~mfem+cuda cuda_arch=70",
      "%xl@16.1.1.11~openmp+mfem+cuda cuda_arch=70"]
  - [arch=linux-rhel7-ppc64le]
- matrix:
  - [axom~cuda+devtools+c2c]
  - ["%clang@9.0.0~openmp+mfem",  "%gcc@7.3.1~mfem", "%xl@16.1.1.12~openmp+mfem"]
  - [arch=linux-rhel7-ppc64le]
```

I doubt that's actually the best way to express that information, but that's an option.

You can also put matrices inside the `definitions` section, and you can put matrices inside of matrices with references to named lists.

cyrush:
Sounds like we are going to need tensor support. (J/K)

PR #83, opened 2022-06-17
Flush print

adrienbernede:
I am flushing all the print() calls and the be logs are now ordered correctly !

Note : Requires python 3.3+

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -837,6 +837,7 @@ class SpackEnv(UberEnv):
         spec_cmd = "spack/bin/spack spec {0} {1}{2}".format(options,self.pkg_name,self.opts["spec"])
 
         res, out = sexe(spec_cmd, ret_output=True, echo=True)
+        print(out, flush=True)
 
         #Check if spec is already installed
         for line in out.split("\n"):
--- a/uberenv.py
+++ b/uberenv.py
@@ -70,7 +70,7 @@ from os.path import join as pjoin
 from os.path import abspath as pabs
 
 
-def sexe(cmd,ret_output=True,echo=False):
+def sexe(cmd,ret_output=False,echo=False):
     """ Helper for executing shell commands. """
     if echo:
         print("[exe: {0}]".format(cmd), flush=True)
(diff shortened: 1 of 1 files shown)


white238 (approved):

adrienbernede:
I know, we can all strike out a skill from our resume...

white238:
I am a little worried that people will have to remember to put `flush=True` on each print. `-u` is supposed to solve this but if people run `python3 uberenv.py` then it doesn't obey `-u` at the top of the file.  A global setting for this would be preferable but the workarounds I see online are not exactly clean.

adrienbernede:
The GitHub CI is failing because it runs python2:
Screen Shot 2022-06-19 at 10 15 16 PM

Do we really want to maintain python2 compatibility, or does the CI config need an update?

adrienbernede:
@white238 I wasn’t able to get the "global flush" to work yet. I need to switch task, but I am willing to try again.
In the meantime, this implementation works. I vote for a merge.

adrienbernede:
I couldn’t resist a new attempt, and it worked. So here we are: global flush applying to all print calls.

Still requires python3, hence failures in CI.

adrienbernede:
@cyrush, Chris and I agreed that we could drop python2 support. This is what I did in the CI config so that the test can pass.
I’d like to have your approval before I merge this.

cyrush (approved):

white238 (approved):
Thanks @adrienbernede !

docs/sphinx/index.rst
```diff
@@ -81,6 +81,9 @@ Uberenv can also be used as a submodule of the user project, where one must prov
 additionally provide the entries ``spack_configs_path`` and ``spack_packages_path``.
 See :ref:`project_configuration` for more details.
 
+.. Note::
+   Uberenv requires python 3.3 or above.
```
kennyweiss, line 85: This is noteworthy! Please add it to the `RELEASE-NOTES` too.

kennyweiss (approved):
Thanks @adrienbernede !

PR #84, opened 2022-07-11
Woptim/quotes for specs

adrienbernede:
In this PR I add single quotes around Spack specs to work around an issue in Spack with complex flags parsing.
(see https://github.com/spack/spack/issues/31214).

This issue arise when using Spack newer than https://github.com/spack/spack/commit/0c7fd9bd8ca0e38cb5a497acdfc40f51170a07f8.

There is one caveat: With this change the spec in CI should always use double quotes, and flag quotes should be protected:

```
-    SPEC: "~fortran %clang@9.0.0+libcpp cxxflags='-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0' cflags='-DGTEST_HAS_CXXABI_H_=0'"
+    SPEC: "~fortran %clang@9.0.0+libcpp cxxflags=\"-stdlib=libc++ -DGTEST_HAS_CXXABI_H_=0\" cflags=\"-DGTEST_HAS_CXXABI_H_=0\""
```

So this may be breaking for some projects, but easily fixable.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -680,7 +680,7 @@ class SpackEnv(UberEnv):
 
     # Extract the first line of the full spec
     def read_spack_full_spec(self,pkg_name,spec):
-        res, out = sexe("spack/bin/spack spec '" + pkg_name + " " + spec "'", ret_output=True)
+        res, out = sexe("spack/bin/spack spec '" + pkg_name + " " + spec + "'", ret_output=True)
         for l in out.split("\n"):
             if l.startswith(pkg_name) and l.count("@") > 0 and l.count("arch=") > 0:
                 return l.strip()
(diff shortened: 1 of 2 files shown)


uberenv.py
```diff
@@ -680,7 +680,7 @@ def find_spack_pkg_path(self, pkg_name, spec = ""):
 
     # Extract the first line of the full spec
     def read_spack_full_spec(self,pkg_name,spec):
-        res, out = sexe("spack/bin/spack spec " + pkg_name + " " + spec, ret_output=True)
+        res, out = sexe("spack/bin/spack spec '" + pkg_name + " " + spec + "'", ret_output=True)
```
kennyweiss, line 683: Would this be easier to read if it used `String.format()` instead of concatenation?

adrienbernede: Probably, but in hacking mode, I just add the minimum number of symbols required :)
I can change that tomorrow.

adrienbernede:
Changes validated in Umpire, release notes updated accordingly.

kennyweiss (approved):
Thanks @adrienbernede !

PR #86, opened 2022-07-12
`--fresh` vs `--reuse` cleanup

### LINKED ISSUE #85: Spack now defaults to `--reuse` we need to add `--fresh`
Spack has started to find packages in other directories when `--reuse` is not on.  We need to add `--fresh` due to a default being flipped to stop this from happening.

white238:
Check for the existence of `--fresh` and `--reuse` before using them.  Default to `--fresh` when `--reuse` is not specified by user.

Fixes #85 .

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -585,22 +585,28 @@ class SpackEnv(UberEnv):
 
         print("[spack spec: {0}]".format(self.opts["spec"]))
 
-        # Use '--fresh' argument if it exists and if not using '--reuse'
+        # List of concretizer options not in all versions of spack
         # (to be checked if it exists after cloning spack)
         self.fresh_exists = False
+        self.reuse_exists = False
 
     def check_for_fresh(self):
         spack_dir = self.dest_spack
         cmd = pjoin(spack_dir,"bin","spack")
         cmd += " help install"
-        print("[Checking for --fresh...]")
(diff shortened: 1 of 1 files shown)


white238:
I verified this worked in : https://github.com/LLNL/serac/pull/751

Notice the no longer overlapping `TPL_ROOT` directories.. woo~

white238:
This is ready for review.  Thanks for the patience!

uberenv.py
```diff
@@ -585,6 +585,33 @@ def __init__(self, opts, extra_opts):
 
         print("[spack spec: {0}]".format(self.opts["spec"]))
 
+        # List of concretizer options not in all versions of spack
+        # (to be checked if it exists after cloning spack)
+        self.fresh_exists = False
+        self.reuse_exists = False
+
+    def check_for_fresh(self):
```
kennyweiss, line 593: Suggestion: Please consider renaming this function to:
```suggestion
    def check_concretizer_opts(self):
```

kennyweiss (approved):
Thanks @white238

uberenv.py
white238: Good catch... pushing in a bit

adrienbernede (approved):
All good, tested on Umpire.

PR #92, opened 2022-11-04
Skip setup option

chapman39:
New uberenv option `--skip-setup`, which skips cloning Spack, patching, and clean commands. This option is to support close-side builds (so they won't be able to clone Spack).

How I tested:
`cd .ci/test-project`
`python3 ../../uberenv.py --project-json=install_config.json --setup-only`
`python3 ../../uberenv.py --project-json=install_config.json --skip-setup`


white238:
Can you add a CI job to test this? Maybe one that does just the setup (check if spack is there and can run maybe a help command) then another one with setup only then skip setup?

white238:
Need release notes and document it in the sphinx docs as well.

.github/workflows/linux.yml
```diff
@@ -85,3 +85,19 @@ jobs:
...
+      run: |
+          sudo apt-get update
+          sudo apt-get install $BASE_PACKAGES
+    - uses: actions/checkout@v3
+    - name: Run Uberenv
+      run: |
+          cp .ci/test-project/packages/magictestlib.tar.gz . && tar -xvf magictestlib.tar.gz && cp src/* .
+          cd .ci/test-project
+          python3 ../../uberenv.py --project-json=install_config.json --spack-config-dir=spack_configs --setup-only
+          python3 ../../uberenv.py --project-json=install_config.json --spack-config-dir=spack_configs --skip-setup
```
white238, line 103: Can you add a basic check that a file exists in the install directory after this?

docs/sphinx/index.rst
```diff
@@ -110,6 +110,8 @@ Uberenv has a few options that allow you to control how dependencies are built:
   ``-k``                  Ignore SSL Errors                                    **False**
   ``--install``           Fully install target, not just dependencies          **False**
   ``--run_tests``         Invoke tests during build and against install        **False**
+  ``--setup-only``        Only download and setup Spack                        **False**
+  ``--skip_setup``        Only install (using pre-existing environment)        **False**
```
white238, line 114:
```suggestion
  ``--skip_setup``        Only install (using pre-setup spack)        **False**
```

white238: environment is a very overloaded word

.github/workflows/linux.yml
chapman39: added a check to the other "full" ci tests as well

.github/workflows/linux.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           cp .ci/test-project/packages/magictestlib.tar.gz . && tar -xvf magictestlib.tar.gz && cp src/* .
           cd .ci/test-project
           python3 ../../uberenv.py --project-json=dev_build_config.json --spack-config-dir=spack_configs
+          ./uberenv_libs/spack/opt/spack/linux-*/gcc-*/magictestlib_cached-1.0.0-*/bin/uberenv_conduit_hello
```
white238, line 43: 👍

uberenv.py
```diff
@@ -246,6 +246,13 @@ def parse_args():
                       default=False,
                       help="Only download and setup Spack. No further Spack command will be run.")
 
+    # option to stop after spack download and setup
```
white238, line 249:
```suggestion
    # option to start after spack download and setup
```

white238 (approved):

docs/sphinx/index.rst
```diff
@@ -110,6 +110,8 @@ Uberenv has a few options that allow you to control how dependencies are built:
   ``-k``                  Ignore SSL Errors                                    **False**
   ``--install``           Fully install target, not just dependencies          **False**
   ``--run_tests``         Invoke tests during build and against install        **False**
+  ``--setup-only``        Only download and setup Spack                        **False**
+  ``--skip_setup``        Only install (using pre-setup Spack)                 **False**
```
adrienbernede, line 114: I think we should use either "-" or "_" in option names, for consistency and ease of use. I suggest "-".

uberenv.py
```diff
@@ -105,7 +105,7 @@ def parse_args():
                       default=None,
                       help="destination directory")
 
-    # what compiler to use
+    # what compiler to use and other package configurations
```
adrienbernede, line 108: We could also change the option description which is also inaccurate.
Maybe something like "Standard spack spec, but package name is optional".

adrienbernede (dismissed):
Minor changes requested. Please ping me when done.

docs/sphinx/index.rst
chapman39: turns out in `uberenv.py` itself i had it with a "-" but in the docs it was "_", for `skip-setup`.

theres also `--run_tests` which uses an underscore. i would change it to use a hyphen but i think that would cause issues for users.

uberenv.py
chapman39: the description is now "standard spack spec"

RELEASE-NOTES.md
```diff
@@ -10,6 +10,8 @@ The Uberenv project release numbers follow [Semantic Versioning](http://semver.o
 ## Unreleased
 
 ### Added
+- Adds `--skip-setup` option, which skips installing and setting up Spack and installs only
+  (using pre-setup Spack). Useful for close-side builds.
```
cyrush, line 14:
```suggestion
  (using pre-setup Spack). Useful for air-gapped builds.
```

cyrush (approved):

PR #94, opened 2022-11-12
allow deprecated pkgs to build by setting config opt during patch

cyrush:


cyrush:
This setting is needed because without it (or related command line option `--deprecated`) spack install will error out if you try to build a package marked deprecated. 

I hit this while moving to a newer version of spack for conduit, where we have CI tests that test python2 support.

I also tried adding `--deprecated`, but this is simpler than plumbing that into uberenv spack commands

adrienbernede:
I understand that this is way easier than adding an option. But I’d like uberenv to behave as close as possible to Spack by default. Here, it would not be possible to run in "vanilla Spack" mode.

Instead of the painful "plumbing into uberenv spack commands", maybe you could encapsulate you single-line-change in a conditional controlled by a `--deprecated` option ?

What do you think ?

cyrush:
Yes, I will add a specific Uberenv option and connect that to the install command. I tried to add a catch all “spack extra args” option, but that was hard to plumb. Specific option would work well, I’ll move to that.

-Cyrus
________________________________

cyrush:
@adrienbernede I took an initial pass & test later this week.  Not sure if we also need `--deprecated` in the info/status calls.

kennyweiss (approved):
Thanks @cyrush

A more general option might be to add a way to pass in additional options to the spack command line(s), rather than a dedicated `--deprecated` flag, but this one's fine with me.

cyrush:
@kennyweiss I actually tried first - before the config add solution. The current solution is the third one.

To make it general, we would have to have pre and post options (reflecting before the spack sub command, or after). And in the `--deprecated` case we would not want them on all invocations of spack to use the args, so it really is better to have a targeted.

adrienbernede:
This looks good to me in terms of strategy. Don’t we need "deprecated" in the spec command too ? (to print the same spec than the one finally installed ?).

cyrush:
When I use the `--spack--allow-deprected` option, `--prefix` is ignored. 

```
python3 scripts/uberenv/uberenv.py -k --spack-allow-deprecated --prefix="ul_t"
...
[info: destination '/Users/harrison37/Work/github/llnl/conduit/uberenv_libs' already exists]
```

```
python3 scripts/uberenv/uberenv.py -k  --prefix="ul_t"
...
[info: destination '/Users/harrison37/Work/github/llnl/conduit/ul_t' already exists]
```

I don't see how adding an option like this would undermine other options - what in the world am I doing wrong?

cyrush:
I see now :-(

white238:
> I see now :-(

My guess it was parsing the next argument as a string to that one...

```
    parser.add_option("--spack-allow-deprecated",
                      dest="spack_allow_deprecated",
                      action="store_true"
                      default=False,
                      help="add --deprecated to spack install commands")
```

Pretty sure all of these options aren't fully correct.

cyrush:
Update; `--deprecated` does not need to be passed to the spec command, in fact it causes an error:

```
[exe: spack/bin/spack  spec --fresh --install-status --very-long --deprecated  'uberenv-conduit@develop^python@2.7 ']
==> Error: unrecognized arguments: --deprecated
```

uberenv.py
```diff
@@ -174,6 +174,12 @@ def parse_args():
                       dest="spack_debug",
                       default=None,
                       help="add debug option to spack spec/install commands")
+    # spack allow deprecated packages
```
white238, line 177: Can you fix the `--spack-debug` command above this one?

white238 (approved):

cyrush:
@white238 done.

adrienbernede (approved):
Thank you for the fixes, and trying the spec option.

Issue #93, opened 2022-11-04
Move to Spack Environments


white238:
We need to move away from the old Spack usage of a split packages/compilers file (I don't think there is a word for it) and to Spack Environments.

I had a long conversation with @becker33 and this was his recommendation to convert:

```
# Translation of dev-build -u X in uberenv-speak
spack env create env /path/to/environment/file
spack -e env add $spec
# dev-build mode command
spack -e env develop -u final_phase--no-clone --path $location_of_axom $spec  # remove -u final_phase for install
# install and uberenv-pkg modes command
spack -e env install -u final_phase --no-add $spec  # remove -u final_phase for full install
```

Note this example is for `dev-build with a final phase` mode but has alterations for `install` and `dev-build with no final phase`.

white238:
I am pretty sure there should be a `spack env create env /path/to/environment/file` before the `add` command.

white238:
TODO:

- [ ] Move unified spack commands out of #75 into new PR, merge
- [ ] Have uberenv install host-config into uberenv prefix
- [ ] Create new Env PR and move current `uberenv.py` to `pre_env_uberenv.py`

Cyrus concerns (may not need to be fixed before):
https://github.com/LLNL/uberenv/issues/44
https://github.com/LLNL/uberenv/pull/46#issuecomment-796968130

PR #95, opened 2022-12-08
Unify spack command into separate function

chapman39:
In preparation for moving to spack environments mentioned in issue https://github.com/LLNL/uberenv/issues/93 and was originally from PR https://github.com/LLNL/uberenv/pull/75.

 - Instead of writing the spack executable path completely each time, calls `spack_exe_path()`
 - Prints spack version in `show_info()`
 - Additionally prints hash on error in `find_spack_pkg_path_from_hash()`
 - Additionally prints spec on error in `find_spack_pkg_path()`
 - Fixes bug where sometimes the spack package hash has a leading space, which causes it not to be found in `find_spack_pkg_path_from_hash()`

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -64,6 +64,7 @@ import glob
 import re
 
 from optparse import OptionParser
+from distutils.version import LooseVersion
 from functools import partial
 
 from os import environ as env
@@ -347,6 +348,16 @@ class UberEnv():
         print("[uberenv command line options: ")
         pretty_print_dictionary(self.opts)
         print("]")
+    
+    ###########################
+    # basic spack helpers
+    ###########################
+    def spack_exe_path(self):
+        return pjoin(self.dest_dir,"spack/bin/spack")
+
+    def spack_version(self):
+        res, out = sexe('{0} --version'.format(self.spack_exe_path()), ret_output=True)
+        return LooseVersion(out)
 
     def setup_paths_and_dirs(self):
         self.uberenv_path = uberenv_script_dir()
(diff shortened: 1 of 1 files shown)


uberenv.py
```diff
@@ -347,6 +348,16 @@ def __init__(self, opts, extra_opts):
...
         pretty_print_dictionary(self.opts)
         print("]")
+    
+    ###########################
+    # basic spack helpers
+    ###########################
+    def spack_exe_path(self):
+        return pjoin(self.dest_dir,"spack/bin/spack")
+
+    def spack_version(self):
```
white238, line 358: Were you planning on calling this somewhere?

chapman39: `spack_version`? This was from Cyrus's PR. I wasn't planning on using it in the PR, so I can remove it.

cyrush (approved):
Looks good!

kennyweiss (approved):
Looks good @chapman39 

Suggestion for the future: Instead of calling the shell execute (`sexe`) command with a first argument of `self.spack_exe_path` would it make sense to have a function called `spack_command` where we pass in the command and options more explicitly?

E.g. instead of 
```python
sexe('{0} find -p /{1}, self.spack_exe_path(), pkg_hash)
```
consider:
```python
spack_command("find", "-p /{0}".format(pkg_hash))
```

This might make the code's intent more obvious to readers.

It might also work well with environments since we can add an optional third parameter that passes in the environment and/or uses the default one.

cyrush:
> Looks good @chapman39
> 
> Suggestion for the future: Instead of calling the shell execute (`sexe`) command with a first argument of `self.spack_exe_path` would it make sense to have a function called `spack_command` where we pass in the command and options more explicitly?
> 
> E.g. instead of
> 
> ```python
> sexe('{0} find -p /{1}, self.spack_exe_path(), pkg_hash)
> ```
> 
> consider:
> 
> ```python
> spack_command("find", "-p /{0}".format(pkg_hash))
> ```
> 
> This might make the code's intent more obvious to readers.
> 
> It might also work well with environments since we can add an optional third parameter that passes in the environment and/or uses the default one.

Note on this:  There are cases when we capture output from spack command and cases where we simply run the spack command.  

So if we wrap these into a higher level function, we also need to make sure to wrap the `ret_output` argument semantics as well. 

Handling those cases was one reason why I didn't refactor that far when I took my pass at this, but this is a good idea to tackle in the future -- would simplify the code big time.

white238 (approved):

PR #96, opened 2022-12-17
Spack Environments

### LINKED ISSUE #93: Move to Spack Environments

### LINKED ISSUE #98: Adding a spack package repo instead of copying packages over to spack's builtin location
In SpackEnv's patch function, we manually copy all of the packages in to spack's builtin package repo, but it would be better to get  the location of the project's package repo (in something like `--spack-package-repo-dir`), and do `spack repo add <spack_package_repo_dir>` after creating the spack environment.

This would also involve creating a `repo.yaml` for the ci test project.

I was thinking about doing this after the environment's PR https://github.com/LLNL/uberenv/pull/96

https://spack.readthedocs.io/en/latest/repositories.html

chapman39:
### Changes that will affect users
- Instead of having `compilers.yaml` and `packages.yaml`, they're both put into a single `spack.yaml`
- `--spack-config-dir` option is replaced by `--spack-env-file`, which is the path to a `spack.yaml` or `spack.lock`
- `--spack-env-name` is a new option that names the environment (which is always placed in the destination dir)
- `spack-activate` is no longer an option and is replaced by views

### File Organization Changes
- adds `.ci/test-project/spack_configs` directory with different `spack.yaml`'s based on system
- adds `.ci/test-project/uberenv_configs` directory for each different mode to test (install, dev-build, uberenv-pkg)
- adds `deprecated_uberenv.py`, which is the original `uberenv.py` before spack environments

### Uberenv.py Changes
- If `--spack-env-file` is not specified, `uberenv.py` will try to find it based on the `spack_config_path` and platform
- Moves "spack helper files" to `SpackEnv` class
- Replaces `spack_exe_path()` with `spack_exe()`, which has an option to include Spack's environment `-e` option (default).
- `SpackEnv` has a new function,`create_spack_env()`, that creates a spack environment, adds the package to the environment, and concretizes the environment
- Creates `pkg_name_with_spec` variable to `SpackEnv` class, which is the combination of package name and spec
- Simplifies `SpackEnv`'s `install()`. Every build mode now uses `spack install`. Removes `spack activate` sections.

Fixes #93
Fixes #98


.ci/test-project/spack_configs/linux_ubuntu_22/spack.yaml
```diff
@@ -0,0 +1,18 @@
+# Spack env file for ubuntu 20
```
white238, line 1:
```suggestion
# Spack env file for ubuntu 22
```

docs/sphinx/index.rst
```diff
@@ -161,13 +162,8 @@ Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_configs
...
 * ``{spack_configs_path}`` can be specified in the json config file.
 
-You may instead use the **--spack-config-dir** option to enforce the use of a specific directory. As long as it provides Uberenv with the yaml files to use with Spack.
-See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_ and
-`Spack System Packages <http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>`_ documentation for details.
-
-.. note::
-    The bootstrapping process ignores ``~/.spack/compilers.yaml`` to avoid conflicts
-    and surprises from a user's specific Spack settings on HPC platforms.
+You may instead use the **--spack-env-file** option to enforce the use of a specific directory. As long as it provides Uberenv with the yaml files to use with Spack.
```
white238, line 165:
```suggestion
You may instead use the **--spack-env-file** option to enforce the use of a specific directory, as long as it provides Uberenv with the yaml files to use with Spack.
```

.ci/test-project/spack_configs/linux_ubuntu_22/spack.yaml
chapman39: thanks for catching this

chapman39:
Decided to remove the spack activate sections, given its not even in the latest Spack documentation. I can re-add, if wanted.
https://spack.readthedocs.io/en/latest/command_index.html

.ci/test-project/spack_configs/darwin/spack.yaml
```diff
@@ -0,0 +1,18 @@
+# Spack env file for mac
+spack:
+  packages:
```
white238, line 3: I assume not listing any compilers here allows spack to find something?

.ci/test-project/spack_configs/linux_ubuntu_22/spack.yaml
```diff
@@ -0,0 +1,18 @@
+# Spack env file for ubuntu 22
+spack:
+  packages:
+    cmake:
+      externals:
+      - spec: cmake@3.14.5
```
white238, line 6: Can you verify this cmake version?

deprecated_uberenv.py
```diff
@@ -0,0 +1,1323 @@
...
+# DAMAGES  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+###############################################################################
+
+"""
```
white238, line 47: Can you add a large note here that says why this is deprecated and to favor the non-deprecated version?

.ci/test-project/spack_configs/darwin/spack.yaml
cyrush: in the past we have had to list macOS compilers, but it could have be necessary b/c we were overriding defaults.

docs/sphinx/index.rst
```diff
@@ -161,13 +162,8 @@ Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_configs
...
 * ``{spack_configs_path}`` can be specified in the json config file.
 
-You may instead use the **--spack-config-dir** option to enforce the use of a specific directory. As long as it provides Uberenv with the yaml files to use with Spack.
-See the `Spack Compiler Configuration <http://spack.readthedocs.io/en/latest/getting_started.html#manual-compiler-configuration>`_ and
-`Spack System Packages <http://spack.readthedocs.io/en/latest/getting_started.html#system-packages>`_ documentation for details.
-
-.. note::
-    The bootstrapping process ignores ``~/.spack/compilers.yaml`` to avoid conflicts
-    and surprises from a user's specific Spack settings on HPC platforms.
+You may instead use the **--spack-env-file** option to enforce the use of a specific directory, as long as it provides Uberenv with the yaml files to use with Spack.
```
white238, line 165:
```suggestion
You may instead use the **--spack-env-file** option to enforce the use of a specific Spack Environments file.
```

docs/sphinx/index.rst
```diff
@@ -161,13 +162,8 @@ Uberenv looks for configuration yaml files under ``scripts/uberenv/spack_configs
...
+See the `Environments (spack.yaml) <https://spack.readthedocs.io/en/latest/environments.html>`_ documentation for details.
```
white238, line 166:
```suggestion
See the `Spack Environments (spack.yaml) <https://spack.readthedocs.io/en/latest/environments.html>`_ documentation for details.
```

uberenv.py
```diff
@@ -262,6 +255,17 @@ def parse_args():
                       default=False,
                       help="Only install (using pre-setup Spack).")
 
+    # spack environment directory
+    parser.add_option("--spack-env",
```
white238, line 259: This option is going to confuse people. We need to think of a better name.  Thoughts, @kennyweiss ?

uberenv.py
```diff
@@ -320,7 +320,7 @@ def find_project_config(opts):
                 return project_json_file
             else:
                 lookup_path = pabs(os.path.join(lookup_path, os.pardir))
-    print("ERROR: No configuration json file found")
+    print("ERROR: No uberenv configuration json file found")
```
white238, line 323: 👍

uberenv.py
```diff
@@ -678,14 +682,37 @@ def setup_paths_and_dirs(self):
...
+
+        # Setup path of Spack environment configuration file if not specified on command line
+        # Check under spack_config_path -> detected platform -> spack.yaml/ .lock
+        if self.opts["spack_env_file"] is None:
+            # Check if platform is detected
             uberenv_plat = self.detect_platform()
-            if uberenv_plat is not None:
-                self.spack_config_dir = pabs(pjoin(spack_configs_path,uberenv_plat))
+            if uberenv_plat is None:
+                print("[ERROR: could not detect platform. (Supply Spack environment file in command line with --spack-env-file]")
```
white238, line 698:
```suggestion
                print("[ERROR: Could not detect platform. Supply Spack environment file in command line with --spack-env-file=/path/to/spack.yaml")
```

uberenv.py
```diff
@@ -678,14 +682,37 @@ def setup_paths_and_dirs(self):
...
-        # Test if the override option was used (--spack-config-dir)
-        self.spack_config_dir = self.opts["spack_config_dir"]
-        if self.spack_config_dir is None:
-            # If command line option is not used, search for platform under
-            # given directory
+        # Set spack_env to absolute path and (if exists) check validity
+        if self.opts["spack_env"] is not None:
+            self.spack_env = pabs(self.opts["spack_env"])
+            if os.path.exists(self.spack_env) and not os.path.isdir(self.spack_env):
+                print("[ERROR: invalid spack env directory: {0} ]".format(self.spack_env))
```
white238, line 689:
```suggestion
                print("[ERROR: Invalid Spack environment directory given. File found instead: {0} ]".format(self.spack_env))
```

uberenv.py
```diff
@@ -678,14 +682,37 @@ def setup_paths_and_dirs(self):
...
+            # Check if a path to an init file is located
+            self.spack_env_file = pabs(pjoin(spack_configs_path, uberenv_plat))
+            spack_env_yaml = pjoin(self.spack_env_file, "spack.yaml")
+            spack_env_lock = pjoin(self.spack_env_file, "spack.lock")
+            if os.path.exists(spack_env_yaml):
+                self.spack_env_file = spack_env_yaml
+            elif os.path.exists(spack_env_lock):
+                self.spack_env_file = spack_env_lock
+            else:
+                print("[ERROR: could not find spack env file path Spack environment file (e.g. spack.yaml) under: {0}]".format(self.spack_env_file))
```
white238, line 710:
```suggestion
                print("[ERROR: Could not find Spack environment file path Spack environment file (e.g. spack.yaml) under: {0}]".format(self.spack_env_file))
```

uberenv.py
```diff
@@ -903,12 +923,13 @@ def show_info(self):
         options = ""
         options = self.add_concretizer_opts(options)
         options += "--install-status --very-long"
-        spec_cmd = "{0} {1}spec {2} '{3}{4}'".format(self.spack_exe_path(),debug,options,self.pkg_name,self.opts["spec"])
+        #spec_cmd = "{0} {1}spec {2} {3}".format(self.spack_exe(),debug,options,self.pkg_name_with_spec)
```
white238, line 926: Remove dead code.

.ci/test-project/spack_configs/darwin/spack.yaml
white238: I figured that was true. I assume Spack just goes with the system gnu. Maybe its better to just leave it blank and let Spack handle it.

chapman39: Added the comment 👍

uberenv.py
chapman39: Maybe `spack-env-name` ?

deprecated_uberenv.py
chapman39: Also updated file name on line 48. Added note starting on line 53.

uberenv.py
chapman39: LiDO, for another reference, calls the spack environment directory `venv_dir` and the yaml file defining the spack environment `manifest_file`. Maybe something like that would be good too.

`spack_env -> venv_dir : Spack environment directory`
`spack_env_file -> manifest_file : YAML file defining spack environment`

cyrush: How is this used along with `--prefix`?  We need an extra name?

cyrush: If we do need to control the name, maybe: `--spack-env-dir-name`

.ci/test-project/spack_configs/linux_ubuntu_22/spack.yaml
chapman39: it seems to find it fine here: https://github.com/LLNL/uberenv/actions/runs/3953297149/jobs/6769414659#step:4:114

`[+] /usr (external cmake-3.14.5-ne5mma6cz46xq6alhyop7qjelao6b55r)`

white238: I am pretty sure it just takes what version you said on line 6 and doesn't actually verify the version number.

You either have to add a `cmake --version` to the CI script and look at the output or download the docker image and run the command yourself inside of it.

Looking at this page seems to imply it is 3.22:

https://packages.ubuntu.com/search?keywords=cmake

uberenv.py
white238: `--prefix` is where uberenv does all it's work (downloads spack, creates logs, also where axom/serac's install prefix lives at this level though this is controlled by the spack config.yaml).

`--spack-env-name` (or whatever) is the name of the environment but Spack also creates a directory that holds a lot of information (logs, a spack.yaml, a package repository, bunch of metadata, etc). More info here:

https://spack.readthedocs.io/en/latest/environments.html#creating-a-named-environment

.ci/test-project/spack_configs/linux_ubuntu_22/spack.yaml
chapman39: looks like its 3.25.2. ill update that, thanks. 

https://github.com/LLNL/uberenv/actions/runs/4108316351/jobs/7088871940#step:2:127

uberenv.py
cyrush: I would say `--spack-env-name` sounds good.

Does it have a default (in case you don't care)?

white238: It defaults to `spack_env`

cyrush:
Q: Does this test spack views (or whatever the env construct is) that replaces the old python activate command?

white238:
> Q: Does this test spack views (or whatever the env construct is) that replaces the old python activate command?

He removed the `spack activate` capability (because its broken and/or gone). Spack views will be a follow up PR.

uberenv.py
chapman39: I'll change it to `--spack-env-name`, then

kennyweiss (approved):
Thanks @chapman39!

uberenv.py
```diff
@@ -262,6 +255,17 @@ def parse_args():
                       default=False,
                       help="Only install (using pre-setup Spack).")
 
+    # Spack Environment directory
+    parser.add_option("--spack-env-name",
+                      dest="spack_env_name",
+                      default="spack_env",
+                      help="The created Spack Environment directory. Will create directory automatically if it doesn't exist.")
```
white238, line 262:
```suggestion
    # Spack Environment name
    parser.add_option("--spack-env-name",
                      dest="spack_env_name",
                      default="spack_env",
                      help="The name of the created Spack Environment.")
```

white238: I think it would be helpful to not default to `spack_env` but to actually `spack_env + <time date stamp>`. That way if you run uberenv times, you will not override the same environment file. This would be helpful for debugging after several runs.

cyrush: That would break repeated calls to uberenv however (multiple calls iterating towards success, like if a download fails)?

white238: That's a distinct possibility. Lets just leave it out of this PR and maybe try it afterwards as a targeted PR. I am not sure, without testing personally, what would happen if you just called uberenv multiple times on the same spec.

cyrush: To be clear - repeated calls to the same build is something I do often with the current uberenv during development.

kennyweiss: I do that too. 

How about an extra option to append the date/time which is off by default, but could be added to the uberenv-config file?

white238: I am fine with just leaving it a generic name like it is now. It was just a random thought that clearly everyone hates. <cries>

chapman39: did suggestion and updated docs as well

uberenv.py
```diff
@@ -262,6 +255,17 @@ def parse_args():
...
+
+    # Spack Environment file
+    parser.add_option("--spack-env-file",
+                      dest="spack_env_file",
+                      default=None,
+                      help="Path to Spack Environment File (e.g. spack.yaml or spack.lock)")
```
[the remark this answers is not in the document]
chapman39: i've updated the comments to be consistently "name". but you bring up a good point that it is still confusing. it doesnt help that the spack docs for it is equally as unclear. https://spack.readthedocs.io/en/latest/command_index.html#spack-env-create

maybe `--spack-env-file` can be `--spack-env-init-file`? or `--spack-manifest-or-lock`

personally i think its ok as it is but willing to change if people think of better options.

uberenv.py
```diff
@@ -870,24 +862,50 @@ def patch(self):
...
+            concretizer_cmd = "{0} config --scope defaults add config:concretizer:clingo".format(self.spack_exe(use_spack_env=False))
             res = sexe(concretizer_cmd, echo=True)
             if res != 0:
                 print("[ERROR: Failed to update spack configuration to use new concretizer]")
                 sys.exit(-1)
 
+    def create_spack_env(self):
+        # Create Spack Environment
+        print("[creating spack env]")
+        spack_create_cmd = "{0} env create -d {1} {2}".format(self.spack_exe(use_spack_env=False),
```
white238, line 874: The `-d` option should be inside the prefix. This ends up wherever you run uberenv from, which for me was inside Axom's source repo.

chapman39: spack env will now be created in prefix. docs also updated.

uberenv.py
```diff
@@ -262,6 +255,17 @@ def parse_args():
                       default=False,
                       help="Only install (using pre-setup Spack).")
 
+    # Spack Environment name
+    parser.add_option("--spack-env-name",
+                      dest="spack_env_name",
+                      default="spack_env",
+                      help="The name of the Spack Environment, which will be created in destination directory.")
```
white238, line 262:
```suggestion
                      help="The name of the Spack Environment, which will be created in prefix directory.")
```

uberenv.py
```diff
@@ -262,6 +255,17 @@ def parse_args():
...
+
+    # Spack Environment file
+    parser.add_option("--spack-env-file",
+                      dest="spack_env_file",
+                      default=None,
+                      help="Path to Spack Environment File (e.g. spack.yaml or spack.lock)")
```
white238, line 268:
```suggestion
                      help="Path to Spack Environment file (e.g. spack.yaml or spack.lock)")
```

uberenv.py
```diff
@@ -676,16 +678,39 @@ def setup_paths_and_dirs(self):
...
+
+        # Setup path of Spack Environment File if not specified on command line
+        # Check under spack_config_path -> detected platform -> spack.yaml/ .lock
+        if self.opts["spack_env_file"] is None:
+            # Check if platform is detected
             uberenv_plat = self.detect_platform()
-            if uberenv_plat is not None:
-                self.spack_config_dir = pabs(pjoin(spack_configs_path,uberenv_plat))
+            if uberenv_plat is None:
+                print("[ERROR: Could not detect platform. Supply Spack Environment File in command line with --spack-env-file=/path/to/spack.yaml")
```
white238, line 696:
```suggestion
                print("[ERROR: Could not detect platform. Supply Spack Environment file in command line with --spack-env-file=/path/to/spack.yaml")
```

uberenv.py
```diff
@@ -676,16 +678,39 @@ def setup_paths_and_dirs(self):
...
+                self.spack_env_file = spack_env_yaml
+            elif os.path.exists(spack_env_lock):
+                self.spack_env_file = spack_env_lock
+            else:
+                print("[ERROR: Could not find Spack Environment File (e.g. spack.yaml) under: {0}]".format(self.spack_env_file))
+                sys.exit(-1)
+        else:
+            self.spack_env_file = pabs(self.opts["spack_env_file"])
+        
+        print("[Spack Environment File: {0}]".format(self.spack_env_file))
```
white238, line 713:
```suggestion
        print("[Spack Environment file: {0}]".format(self.spack_env_file))
```

uberenv.py
```diff
@@ -828,38 +852,6 @@ def patch(self):
         self.disable_spack_config_scopes(spack_dir)
         spack_etc_defaults_dir = pjoin(spack_dir,"etc","spack","defaults")
```
white238, line 854: @cyrush I just want you to be aware that this will no longer hot-copy any files into spack's repo. All of these things can be handled inside of the environment file.

uberenv.py
```diff
@@ -1043,24 +1014,24 @@ def install(self):
...
+                hc_fname = os.path.split(hc_path)[1]
+                if os.path.islink(hc_fname):
+                    os.unlink(hc_fname)
+                print("[copying host config file to {0}]".format(pjoin(self.dest_dir,hc_fname)))
+                sexe("cp {0} {1}".format(hc_path,hc_fname))
+                print("[removing project build directory {0}]".format(pjoin(build_dir)))
+                sexe("rm -rf {0}".format(build_dir))
         else:
-            print("[ERROR: Unsupported build mode {0}]".format(self.build_mode))
+            print("[ERROR: unsupported build mode: {0}]".format(self.build_mode))
```
white238, line 1034:
```suggestion
            print("[ERROR: Unsupported build mode: {0}]".format(self.build_mode))
```

white238 (approved):
Looks great. Sorry for the minor nitpicking at the end.

Thanks, @chapman39 !

uberenv.py
kennyweiss: In that case, do we need to add safeguards/options our environment files to not use the user's home directory (e.g. `~/.spack`) ?

chapman39: I'm confused by this. Can this section be removed, then (in another PR)?

white238: The disabling of the user scope config (inside your home directory) is handled by the function `disable_spack_config_scopes`, which is still there.

chapman39: this made me realize that `spack_etc_defaults_dir` was no longer being used, so i removed it

uberenv.py
```diff
@@ -1268,29 +1238,32 @@ def main():
...
             if not is_windows() and opts["upstream"] is not None:
                 env.use_spack_upstream()
 
             return 0
-    else:
-        os.chdir(env.dest_dir)
 
+    # Create Spack Environment
+    if not is_windows():
+        env.create_spack_env()
```
adrienbernede, line 1266: In "create_spack_env" the environment is also concretized. It means that it’s concretized before the mirrors and upstreams are set, so the concretization cannot use them. Both could contain already installed packages, especially when using `--reuse`.

Setting up the mirrors and upstreams must also be done before looking for already installed spec in show info, for the same reason.

In the end, I think both concretization and checking for already installed package should be done right before the install. That does not apply to "creating the environment" nor "showing information" by the way.

adrienbernede (changes requested):
Sorry I’m late for the party...

uberenv.py
```diff
@@ -826,39 +850,6 @@ def patch(self):
...
-                sexe("cp {0} {1}/".format(compilers_yaml, spack_etc_defaults_dir ), echo=True)
-
-            if os.path.isfile(packages_yaml):
-                sexe("cp {0} {1}/".format(packages_yaml, spack_etc_defaults_dir ), echo=True)
-        else:
-            # let spack try to auto find compilers
-            sexe("{0} compiler find".format(self.spack_exe_path()), echo=True)
 
         # hot-copy our packages into spack
         if len(self.packages_paths) > 0:
```
white238, line 855: This section should be augmented to only copy if the packages directory is not a spack repo. If it is a spack repo it should call `spack repo add /path`.

uberenv.py
chapman39: So you're saying setting up mirrors and upstreams should be done first, and then concretize?

adrienbernede: Yes, move concretize after, or move mirrors and upstreams before, that’s the same :)

uberenv.py
adrienbernede: I was keeping that one for a later issue... 😔

uberenv.py
chapman39: should be fixed in the following commit

@white238 thoughts on creating a mirror CI check in the future?

uberenv.py
chapman39: im having to move this from patch() and into a new function after the spack environment has been created

chapman39: to determine if there is a spack package repo, i scan each package path in `spack_packages_path` and check if there is a `repo.yaml` in the previous directory.

if i should just check `package_source_dir` instead, let me know.

adrienbernede: Actually @white238 I’m really thinking that this should be addressed in a subsequent PR: we don’t have that "feature" yet in Uberenv, and we will only delay the merge of this one. Additionally, my personal take on this is that we could/should _require_ that custom packages be defined in a legit spack repo and definitively stop hot copying packages.

adrienbernede: Scrolling down and I see there is already a commit for it :face_palming:.

white238: I agree we should require it. There have been more times than I care to admit that I editted the package file in the repo and reran the specific spack command only to remember it wasn't gonna get copied.

white238: I would vote for doing it in this PR because we already have several breaking changes. Might as well just do it in this one.

cyrush: As long as this works seamlessly. How confident are we that wont open up a new surface of confusion?

Hot copying is dead simple to understand and one of the reasons that uberenv has been flexible despite a host of changes in spack proper.

white238: I personally think it is only an improvement. I recommend we remove the hot-copying and add an error check for correctness and tells you how to fix it. For example, if there isnt a `repo.yaml` in the correct spot it errors immediately and tells you to create one there.

cyrush: I trust you, but that's also not the question I asked :-).

It would be good to know what the new failure modes are due to this change. 

Documenting those is important b/c it is a big change. 

For example:

We never needed to write a `repo.yaml` before. (More yaml files for the yaml overloads :-))

I am fine with moving forward, but we need be prepared to help folks understand how to move and fix things when things go off the rails b/c we know they will.

chapman39: I agree. I will add a section in the documentation for this

white238:
I enabled the documentation here:

https://uberenv.readthedocs.io/en/feature-chapman39-spack_env/

adrienbernede (approved):
I got this to work without much trouble in Umpire CI. Thank you @chapman39 !

balos1:
So am I missing something, or does this change mean that  all of the configs in https://github.com/LLNL/radiuss-spack-configs need to be changed to spack environment syntax, i.e. combine the compilers.yaml and packages.yaml files?

cyrush:
I believe so - if you update to the latest uberenv, yes. 

If you aren't ready for the change (which will take some time with respect to configs), then you should stay on the older version.

white238:
We tagged the pre-spack environments version here:

https://github.com/LLNL/uberenv/tree/v1.0.0

It is also included (but won't be updated) as `deprecated_uberenv.py` at the base of the source repo.

white238:
We should add a documentation section with some guidance on what to do to convert from the old split file way to the new environment file. It isn't a lot of work but may be confusing.

adrienbernede:
@balos1 I’m always glad to hear someone is using `radiuss-spack-configs`.
I prepared radiuss-spack-configs for the new uberenv in the branch here: https://github.com/LLNL/radiuss-spack-configs/tree/woptim/update-uberenv

This is for testing purpose: we are still making changes to radiuss-spack-configs non-env version, so I need to think about the migration process first. Maintaining both is not satisfying.

PR #108, opened 2023-03-17
Find externals/ compilers

### LINKED ISSUE #104: optparse -> argparse
Optparse is deprecated.

### LINKED ISSUE #105: Add ability to create a useful Spack environment via command line
Proposed workflow:

```
uberenv.py --spack-create-environment /some/path/spack.yaml
```

Which calls the following commands:

```
// setup spack
spack env create -d /some/path/spack.yaml
spack compiler find
spack external find --not-buildable <list of important packages coming from .uberenv_config.json>
```

chapman39:
* Introduces 2 new arguments:
  * `spack-compiler-paths`: Space delimited string of paths for Spack to search for compilers (if spack_setup_environment is True)
  * `spack-externals`: Space delimited string of packages for Spack to search for externals (if spack_setup_environment is True)
* If Uberenv cannot find a `spack.yaml`, it will automatically create one, search for compilers/ externals, and copy it over to `package_source_dir`.
* Uberenv no longer stops if it cannot find a `spack.yaml`
* Using argparse instead of optparse

Misc Notes:
* I've found that `hdf5-1.14.1-2` inconsistently works. It often gets a FetchError, which is why I try to build `@1.14.0` in CI.

Fixes #105 
Fixes #104


chapman39:
@white238 Was wondering about the workflow of `--spack-create-environment` (or maybe `spack-setup-environment` makes more sense?)

Were you thinking the user would run uberenv with `--spack-setup-environment` first, and uberenv would copy the spack.yaml over to a user-specified location (like `<proj>/scripts/spack/configs/<platform>)? And then later on you'd modify the spack.yaml and run uberenv again.

uberenv.py
```diff
@@ -867,6 +875,29 @@ def create_spack_env(self):
...
+        # Find pre-installed compilers and packages and stop uberenv.py
+        if self.opts["spack_setup_environment"]:
+            print("[finding compilers]")
+            spack_compiler_find_cmd = "{0} compiler find".format(self.spack_exe())
+            res_compiler = sexe(spack_compiler_find_cmd, echo=True)
+            if res_compiler != 0:
+                print("[failed to setup environment]")
+                sys.exit(-1)
+
+            if not self.spack_not_buildable_packages is None:
```
white238, line 888: I think we want this to happen even if the list is empty

.ci/test-project/uberenv_configs/install.json
```diff
@@ -5,6 +5,7 @@
 "spack_build_mode" : "install",
 "spack_configs_path": "spack_configs",
 "spack_packages_path": "packages",
+"spack_not_buildable_packages": "cmake pkg-config zlib",
```
white238, line 8: I don't like this name but can't come up with a better one right now. Let me ponder this.

chapman39: spack_packages_to_find ?

chapman39:
i've noticed `spack compiler find` finds way more compilers than users would probably want to find. unlike external find, you cant ask for a specific compiler to look for (e.g. gcc)-- as far as i know

```shell
[exe: /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack/bin/spack -D /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack_env compiler find]
==> Added 61 new compilers to /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack_env/spack.yaml
```

white238:
> i've noticed `spack compiler find` finds way more compilers than users would probably want to find. unlike external find, you cant ask for a specific compiler to look for (e.g. gcc)-- as far as i know
> 
> ```shell
> [exe: /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack/bin/spack -D /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack_env compiler find]
> ==> Added 61 new compilers to /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack_env/spack.yaml
> ```

LC machines are very special that they have a million compilers. I don't think this is going to be true on most peoples machines.

But we might want to add how to specify a specific family, `gcc` vs `clang`

cyrush:
In the past when no config was given, uberenv would  look for compilers and a user could ask for the family they wanted as part of the spec passed via command line args (--spec=`clang`,etc). 

It would be good to know how to do that easily with the env setup.

chapman39:
also, running plain `spack compiler find` on ruby doesn't even find the gcc versions we want. (it only found `gcc@4.8.5`)

the only way i could find the gcc compilers with `compiler find` is including the compiler path directly.
```shell
[meemee@ruby967:test-project]$ spack compiler find /usr/tce/packages/gcc/*
==> Added 7 new compilers to /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack_env/spack.yaml
    gcc@10.2.1  gcc@9.3.1  gcc@8.3.1  gcc@8.1.0  gcc@7.3.0  gcc@7.1.0  gcc@6.1.0
==> Compilers are defined in the following files:
    /usr/WS2/meemee/uberenv/repo/.ci/test-project/uberenv_libs/spack_env/spack.yaml
```

chapman39:
as far as i know `spack compiler find` doesn't have an option to search for a specific family
https://spack.readthedocs.io/en/latest/command_index.html#spack-compiler-find

cyrush:
correct, but there was still an easy path to select from what was found. Seems I restated Chris' suggestion.  Just need to how to select.

white238:
> as far as i know `spack compiler find` doesn't have an option to search for a specific family https://spack.readthedocs.io/en/latest/command_index.html#spack-compiler-find

There is a way to tell it to search a different prefix. My guess is by default it only searches in `/usr` or `/opt`. For LC, it also needs to search in `/usr/tce/packages`.

.ci/test-project/uberenv_configs/install.json
adrienbernede: `spack_externals` ?

adrienbernede:
We can pass a path to the find command, that’s correct.
Also, my may issue with it is that it will not find certain compilers, you would have to load the appropriate module.

So, here is the work around I can suggest:
given a compiler family (gcc), or a compiler module name (gcc/8.3.1), look for the PATH prepend in the module files and use those paths to instruct spack where to find the compilers.

adrienbernede:
I guess since Spack is meant to install a full software stack on a machine from scratch, they can’t assume modules are already set. It is true that this may not be portable to machines where modules are not set, but that could be tested by the script.

chapman39:
I've been having trouble implementing `spack_externals` and `spack_compiler_paths` as a command line argument due to the fact that for some reason when I have spaces in the argument it only takes the first thing.

For example:
`uberenv.py --spack-externals="cmake hdf5 zlib"`, will only take `cmake`.

I believe this could be fixed it we switched from OptParse (which has been deprecated since Python 3.2) to ArgParse.

Alternatively of course, we can delimit the argument without spaces (with `;` or `,`).

white238:
> I've been having trouble implementing `spack_externals` and `spack_compiler_paths` as a command line argument due to the fact that for some reason when I have spaces in the argument it only takes the first thing.
> 
> For example: `uberenv.py --spack-externals="cmake hdf5 zlib"`, will only take `cmake`.
> 
> I believe this could be fixed it we switched from OptParse (which has been deprecated since Python 3.2) to ArgParse.
> 
> Alternatively of course, we can delimit the argument without spaces (with `;` or `,`).

Please upgrade to argparse.

chapman39:
Also, thoughts on new variable names?

- `spack-setup-envionment`: (true/false) Setup Spack's environment by searching for pre-installed packages and compilers
- `spack-compiler-paths`: Space delimited string of paths for Spack to search for compilers (if spack_setup_environment is True)
- `spack-externals`: Space delimited string of packages for Spack to search for externals (if spack_setup_environment is True)

uberenv.py
```diff
@@ -687,50 +720,53 @@ def setup_paths_and_dirs(self):
...
+                    self.spack_env_file = spack_env_yaml
+                elif os.path.exists(spack_env_lock):
+                    self.spack_env_file = spack_env_lock
+                else:
+                    print("[WARNING: Could not find Spack Environment file (e.g. spack.yaml) under: {0}]".format(self.spack_env_file))
+                    self.spack_env_file = None
             else:
-                print("[ERROR: Could not find Spack Environment file (e.g. spack.yaml) under: {0}]".format(self.spack_env_file))
-                sys.exit(-1)
+                print("[WARNING: Could not detect platform. Supply Spack Environment file using the command line argument: --spack-env-file=/path/to/spack.yaml")
```
kennyweiss, line 756: Minor suggestion: 
Instead of the double negative, if might be easier to read by switching the order to
```python
  if uberenv_plat in None:
    print("[Warning: could not detect....")
  else:
    # check if a platform....
```

(Please ignore this suggestion if the warning goes away).

kennyweiss, line 756: If this is no longer an `Error`, and we have potential workarounds, should it still be a `Warning`?

It seems like we handle this case with a message a few lines down.

uberenv.py
```diff
@@ -864,13 +900,48 @@ def patch(self):
...
         sexe(spack_create_cmd, echo=True)
+        
+        # Find pre-installed compilers and packages and stop uberenv.py
+        if self.spack_setup_environment:
+            # Finding compilers
+            print("[finding compilers]")
+            if self.spack_compiler_paths is None:
+                spack_compiler_find_cmd = "{0} compiler find".format(self.spack_exe())
+            else:
+                spack_compiler_find_cmd = "{0} compiler find {1}".format(self.spack_exe(), self.spack_compiler_paths)
```
kennyweiss, line 916: Meta-question (outside the scope of this PR):

Should we have a `spack_command(command, additional_args)` function in uberenv?
That way, this line could be written as something like:
```python
    spack_compiler_find_cmd = spack_command("compiler find", self.spack_compiler_paths)
```

kennyweiss (approved):
Thanks @chapman39 for this nice addition!

We should probably add some new CI tests to ensure this continues to work.

uberenv.py
chapman39: yeah the warning may not be necessary anymore

chapman39: @kennyweiss decided to remove the warning

chapman39:
At some point Spack listed spec versions with an `=` sign, which caused some issues with the `find_spack_pkg_path_from_hash` function. This has been fixed in spack in this PR: https://github.com/spack/spack/pull/37663

uberenv.py
```diff
@@ -242,52 +242,73 @@ def parse_args():
...
     # option to skip spack download and setup
-    parser.add_option("--skip-setup",
+    parser.add_argument("--skip-setup",
                       action="store_true",
                       dest="skip_setup",
                       default=False,
                       help="Only install (using pre-setup Spack).")
 
+    # Spack Environment Setup
+    parser.add_argument("--spack-setup-environment",
```
white238, line 259: This is no longer used and should be removed.

white238:
The documentation needs to be updated

uberenv.py
white238: nm.. i searched for the wrong thing.. carry on.

But shouldn't it not be needed anymore because not providing a spack.yaml is implying this option.

white238:
```
./scripts/uberenv/uberenv.py --prefix=../libs
```

produces the following error:

```
Traceback (most recent call last):
  File "/mnt/c/Users/white238/projects/serac/repo/./scripts/uberenv/uberenv.py", line 1380, in <module>
    sys.exit(main())
  File "/mnt/c/Users/white238/projects/serac/repo/./scripts/uberenv/uberenv.py", line 1319, in main
    env.setup_paths_and_dirs()
  File "/mnt/c/Users/white238/projects/serac/repo/./scripts/uberenv/uberenv.py", line 758, in setup_paths_and_dirs
    if self.spack_env_file is None:
AttributeError: 'SpackEnv' object has no attribute 'spack_env_file'. Did you mean: 'spack_env_name'?
```

white238:
Docs here:

https://uberenv.readthedocs.io/en/feature-chapman39-find_externals_and_compilers/

white238 (approved):

white238:
Thanks @chapman39 !

PR #109, opened 2023-03-24
Guard calling `pip` on machines which don't have internet

white238:
Also fixes some copyrights.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -284,6 +284,17 @@ def parse_args():
             opts["mirror"] = pabs(opts["mirror"])
     return opts, extras
 
+def have_internet(host="llnl.gov", port=80, timeout=3):
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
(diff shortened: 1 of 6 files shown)


uberenv.py
```diff
@@ -284,6 +284,17 @@ def parse_args():
...
     return opts, extras
 
+def have_internet(host="llnl.gov", port=80, timeout=3):
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(3)
+        s.connect((host, port))
+        return True
+    except socket.error as ex:
+        print(ex)
```
kennyweiss, line 294: Should this print out the exception? It seems like the exception is expected on some platforms, so it'd be reasonable to omit the message.

kennyweiss (approved):

uberenv.py
white238: Fixed.

chapman39 (approved):

PR #111, opened 2023-03-29
Remove unneeded spec on install cmd

### LINKED ISSUE #110: Investigate if we are adding the spec unnecessarily to the various install commands
I noticed that we are adding the spec to the Spack environment and then also adding it to the various Spack install commands here:

https://github.com/LLNL/uberenv/blob/main/uberenv.py#L987

I am pretty sure this does not need to happen. This should be enough, but I haven't tested it in all the build modes:

```
spack add foo
spack install
```

chapman39:
Fixes #110

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -984,7 +984,6 @@ class SpackEnv(UberEnv):
             if self.opts["build_jobs"]:
                 install_cmd += "-j {0} ".format(self.opts["build_jobs"])
             
-            install_cmd += self.pkg_name_with_spec
             res = sexe(install_cmd, echo=True)
             if res != 0:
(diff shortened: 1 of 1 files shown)


white238 (approved):

kennyweiss:
@chapman39 -- what does the `install_cmd` look like before/after this change?

kennyweiss (approved):
Thanks @chapman39

chapman39:
@kennyweiss 
> what does the install_cmd look like before/after this change?

before: `spack -D spack_env install --fresh magictestlib@1.0.0%gcc@8.3.1`
after: `spack -D spack_env install --fresh`

It does not need the spec here, since it was already specified with the previous `spack add <spec>`

Issue #87, opened 2022-08-31
add spack debug option


cyrush:
When things go wrong in spack land, the first request from spack folks is the stack trace obtained with:

`spack --debug --stacktrace <command>`

We should have a `spack_debug` option that allows users to easily find this info vs hacking it in.

chapman39:
Currently we only debug a couple of spack commands (`install` and `spec`). Thoughts on adding the following code to the `spack_exe()` function instead? @cyrush @white238 

```
debug = ""
if self.opts["spack_debug"]:
    debug = "--debug --stacktrace "
# Append debug to spack command
```

Otherwise this issue could probably be closed.

PR #112, opened 2023-03-30
Add debug to all Spack cmds

chapman39:
Debug each spack command if specified rather than spec and install

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -774,13 +774,6 @@ class SpackEnv(UberEnv):
         print("[ERROR: Failed to find package from spec named '{0}' with spec '{1}']".format(pkg_name, spec))
         sys.exit(-1)
 
-    # Extract the first line of the full spec
-    def read_spack_full_spec(self,pkg_name,spec):
-        res, out = sexe("{0} spec {2}".format(self.spack_exe(), self.pkg_name_with_spec), ret_output=True)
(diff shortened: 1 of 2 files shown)


cyrush (approved):

uberenv.py
```diff
@@ -770,11 +776,7 @@ def find_spack_pkg_path(self, pkg_name, spec = ""):
 
     # Extract the first line of the full spec
     def read_spack_full_spec(self,pkg_name,spec):
-        debug = ""
-        if self.opts["spack_debug"]:
-            debug = "--debug --stacktrace "
-
-        res, out = sexe("{0} {1}spec {2}".format(self.spack_exe(),debug,self.pkg_name_with_spec), ret_output=True)
+        res, out = sexe("{0} spec {2}".format(self.spack_exe(), self.pkg_name_with_spec), ret_output=True)
```
white238, line 779:
```suggestion
        res, out = sexe("{0} spec {1}".format(self.spack_exe(), self.pkg_name_with_spec), ret_output=True)
```

chapman39: Thanks for catching this. This function is never being called anywhere by the way-- not sure if we need to keep it or not.

white238: If it isn't being called you can just remove it.

chapman39:
Fixes #87

Issue #120, opened 2023-08-30
spack envs + uberenv python issues


cyrush:
Two issues with using uberenv + spack envs that we still need to resolve to restore parity with pre spack env uberenv.

These are both related to the requirement that after we use spack to build, we can develop without extra env vars by just using a spack generated host-config. 

1) 

If you are using python modules for your project, an external python will not produce an env view that has a working python w/ spack built modules. Also, uberenv calling spack external find will almost always find a system python.

Both of these mean a default invocation of uberenv will fail. 

Solution? Make sure uberenv excludes looking for python, or add an option to do so. 

2) 

Chicken vs Egg with views to support python use outside of spack.

Assuming you are not using a system python, at the end of the env build, spack will create view that provides python layout that works with all of your requested modules. 

This is great, however we capture the path to python and write it to our host configs during the spack package build process, where this view does not yet exist. 

It works during spack package build b/c spack will provide the proper env vars etc, but the captured host config python path  will not work. 

Solution? Add an option to modify the final host config post build to inject the proper python path.  
Feels messy, but w/o a fix development using python modules is broken w/o other post spack steps.

cyrush:
Note: multiple excludes must be passed to spack in the following way:
 
```
spack external find --exclude python --exclude hdf5
```

PR #124, opened 2023-09-06
uberenv updates to support dev with python in spack envs

### LINKED ISSUE #123: symlinking host config is broken with envs + extra spec constraints
This logic needs to be  updated:

```
pkg_path =  self.find_spack_pkg_path_from_hash(self.pkg_name, self.spec_hash)
```
returns, `pkg_path`: 

```
{'name': 'uberenv-conduit', 'path': '~adios+doc+mpi+openmp+parmetis+python+silo'}
```

cyrush:
Updates that adjust uberenv spack env setup so that folks can develop using python modules installed to a spack view. 

Addresses most challenges outlined #120.

May still need some cleanup, so comments welcome. 

Details:

- Adds a new step that patches host configs post build to use the paths in the view. 

Uses patterns from `.uberenv_config.json` spack_host_config_patches` entry to drive what is patched:

Example input:
```
  "spack_host_config_patches":
 {
   "PYTHON_EXECUTABLE": "view/python*/bin/python3",
   "SPHINX_EXECUTABLE": "view/python*/bin/sphinx-build"
   }
  ```
Creates a `-patchd.cmake` file in the prefix dir (does not modify original host config)

- adds `--spack-skip-externals` option

This is mostly driven by the limitations of using an external python, but is generally useful for cases where system libs undermine you. 

- patch spack.yaml to create the `view` dir under `spack_env`, instead of a hidden file deeper in the dir hierarchy

This should only happen in cases where spack.yaml lacks view customization. 

- fixes #123

### CODE REVIEW

**Comment by cyrush** [comment]
Looks like i need to fix the view dir patching logic.

File: uberenv.py [hunk 1]
```diff
@@ -788,17 +797,22 @@ def find_spack_pkg_path_from_hash(self, pkg_name, pkg_hash):
             # TODO: at least print a warning when several choices exist. This will
             # pick the first in the list.
             if l.startswith(pkg_name) and len(l.split()) > 1:
-                return {"name": pkg_name, "path": l.split()[-1]}
+                pkg_path = l.split()[-1]
```
cyrush, line 800: part of a fix for #123

uberenv.py
```diff
@@ -788,17 +797,22 @@ def find_spack_pkg_path_from_hash(self, pkg_name, pkg_hash):
...
     def find_spack_pkg_path(self, pkg_name, spec = ""):
-        res, out = sexe("{0} find -p {1}".format(self.spack_exe(),self.pkg_name_with_spec), ret_output = True)
+        requested_pkg_name_with_spec = "'{0}{1}'".format(pkg_name,spec)
+        res, out = sexe("{0} find -p {1}".format(self.spack_exe(), requested_pkg_name_with_spec), ret_output = True)
         for l in out.split("\n"):
             # TODO: at least print a warning when several choices exist. This will
             # pick the first in the list.
             if l.startswith(pkg_name) and len(l.split()) > 1:
-                return {"name": pkg_name, "path": l.split()[-1]}
+                pkg_path = l.split()[-1]
```
cyrush, line 813: part of a fix for #123

cyrush:
I removed the logic that tried to present a better default view by hacking yaml, and instead opt-ed to symlink whatever the view directory is under the prefix. This should be more robust and work with any view setup even if it was customized in an input spack.yaml

cyrush:
I need to test why we can't inspect and symlink the view for the `--install` case. 

```
==> Updating view at /home/runner/work/uberenv/uberenv/.ci/test-project/uberenv_libs/spack_env/.spack-env/view
[exe: /home/runner/work/uberenv/uberenv/.ci/test-project/uberenv_libs/spack/bin/spack -D /home/runner/work/uberenv/uberenv/.ci/test-project/uberenv_libs/spack_env python -c 'env = spack.environment.active_environment();print(env.views["default"].get_projection_for_spec(env.matching_spec("'magictestlib_cached@1.0.0%gcc ^hdf5@1.14.0'")))']
[failed to find spack view info]
```

cyrush:
@white238 @kennyweiss @chapman39  

This branch has been tested and is now in use for both conduit and ascent CI. 
I would like to find a path to get it into uberenv main

uberenv.py
```diff
@@ -894,17 +910,42 @@ def disable_spack_config_scopes(self):
...
+                                                                 pjoin(self.dest_dir,"spack_bootstrap"))
+        res = sexe(bstrp_cmd, echo=True)
 
-    def patch(self):
-        # this is an opportunity to show spack python info post obtaining spack
-        self.print_spack_python_info()
 
+    def patch(self):
         # force spack to use only "defaults" config scope
         self.disable_spack_config_scopes()
```
cyrush, line 939: we have to disable this first and for most, b/c w/o this -- any command may read/write to`~/.spack`

uberenv.py
```diff
@@ -894,17 +910,42 @@ def disable_spack_config_scopes(self):
...
 
+        # set bootstrap dir to avoid conflicts with ~/.spack
+        self.set_spack_bootstrap_dir()
+
+        # this is an opportunity to show spack python info post obtaining spack
+        self.print_spack_python_info()
+
         # setup clingo (unless specified not to)
-        if "spack_setup_clingo" in self.project_args and self.project_args["spack_setup_clingo"] == False:
+        if "spack_setup_clingo" in self.project_args and self.project_args["spack_setup_clingo"].lower() == "false":
```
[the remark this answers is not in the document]
cyrush: this really should be called `spack_boostrap_now()` or something similar -- its not setup clingo any more

PR #127, opened 2023-11-09
Remove usage of distutils

### LINKED ISSUE #99: distutils.version Depreceation Warning
I've noticed this error in CI: `DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.`

https://github.com/LLNL/uberenv/blob/3500404a1d5c23f50a33faa46786c4d8b1194751/uberenv.py#L67

chapman39:
> [distutils](https://docs.python.org/3.10/library/distutils.html#module-distutils) is deprecated with removal planned for Python 3.12

* Printing full output of `spack --version`
* Replace distutils.dir_util.copy_tree with shutil.copytree (Someone should test this on Windows though, since I'm on Mac)

Fixes #99

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -59,12 +59,10 @@ import shutil
 import socket
 import platform
 import json
-import datetime
 import glob
 import re
 import argparse
 
-from distutils.version import LooseVersion
 from functools import partial
 
 from os import environ as env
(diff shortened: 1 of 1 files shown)


uberenv.py
```diff
@@ -59,12 +59,10 @@
 import socket
 import platform
 import json
-import datetime
```
chapman39, line 62: Removed `datetime` as well as it's not being used

white238 (approved):
Thanks!

kennyweiss (approved):
Thanks @chapman39 !

cyrush (approved):

PR #133, opened 2024-05-30
Allow stopping uberenv after Spack environment creation and restart from there

adrienbernede:
### Summary

The primary goal of this PR is to allow users to have Uberenv only setup Spack and create the environment (`--setup-and-env-only`).
Then users can customize the environment and resume the build process (`--skip-setup-and-env`).

This was necessary to add an OCI mirror to the environment. A bug in Spack is preventing me from adding an OCI mirror (using variables to set the authentication) to a Spack.yaml file and then use that file to create a Spack environment. (Issue not opened in Spack yet...).

#### Other changes

- Support `oci` as a prefix for remote mirrors (in addition to `http`.
- Improved message describing options to provide an environment.
- Removed some trailing white spaces (sorry for the meaningless diffs).

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -292,7 +292,7 @@ def parse_args():
     # (it should be relative to where uberenv is run from, so it matches what you expect
     #  from shell completion, etc)
     if not is_windows() and args["mirror"] is not None:
-        if not args["mirror"].startswith("http") and not os.path.isabs(args["mirror"]):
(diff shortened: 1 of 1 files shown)


white238 (approved):

kennyweiss (approved):
Thanks @adrienbernede

chapman39 (approved):

PR #136, opened 2024-09-27
Add autopush option for binary cache mirrors

balos1:
This adds a CLI option to enable Spack's auto push feature for binary mirrors and an option to add trusted gpg keys to the spack keyring.


balos1:
Pinging @adrienbernede .

Issue #137, opened 2024-11-21
Support Spack.yaml files split in several files


adrienbernede:
We moved RADIUSS Spack Configs to Spack environments.
Some RADIUSS Spack Configs users would prefer to have a direct access to the equivalent `packages.yaml` and `compilers.yaml`.

When setting `spack.yaml` to include those files through a relative path, using that environment in uberenv fails because at environment creation, the includes are not copied over at the destination path.

https://github.com/LLNL/uberenv/blob/205672b8b2520d7dc69acefe8738960cd5db0937/uberenv.py#L907

I'd like to allow that behavior by first copying the content of the directory containing the original `spack.yaml` file and then run the spack create directly in the destination directory (although I'm not sure whether that would still be necessary).

@white238 @chapman39 what are your thoughts about this ? Did I miss a requirement ?

PR #138, opened 2024-11-22
Copy env file and directory to support includes

adrienbernede:
Solves #137

This allows `spack.yaml` files to include configuration files as long as they belong to the same directory or sub-directories.

Note:
`spack env create` is not run anymore for existing spack.yaml file. Copy works fine in my tests, but there might be a side effect.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -908,7 +908,7 @@ class SpackEnv(UberEnv):
             res = sexe(spack_create_cmd, echo=True)
         else:
             env_file_dir = os.path.dirname(os.path.abspath(self.spack_env_file))
-            env_copy_cmd = "cp -r {0} {1}".format(env_file_dir, self.spack_env_directory)
+            env_copy_cmd = "cp -r {0}/ {1}".format(env_file_dir, self.spack_env_directory)
             res = sexe(env_copy_cmd, echo=True)
         if res != 0:
             print("[ERROR: Failed to create Spack Environment]")
--- a/uberenv.py
+++ b/uberenv.py
@@ -902,7 +902,6 @@ class SpackEnv(UberEnv):
     def create_spack_env(self):
         # Create Spack Environment
         print("[creating spack env]")
-        res = 0
         if self.spack_env_file is None:
             spack_create_cmd = "{0} env create -d {1}".format(self.spack_exe(use_spack_env=False), self.spack_env_directory)
             res = sexe(spack_create_cmd, echo=True)
--- a/uberenv.py
+++ b/uberenv.py
@@ -902,11 +902,14 @@ class SpackEnv(UberEnv):
     def create_spack_env(self):
         # Create Spack Environment
         print("[creating spack env]")
+        res = 0
         if self.spack_env_file is None:
-            self.spack_env_file = ""
-        spack_create_cmd = "{0} env create -d {1} {2}".format(self.spack_exe(use_spack_env=False),
-            self.spack_env_directory, self.spack_env_file)
-        res = sexe(spack_create_cmd, echo=True)
+            spack_create_cmd = "{0} env create -d {1}".format(self.spack_exe(use_spack_env=False), self.spack_env_directory)
+            res = sexe(spack_create_cmd, echo=True)
+        else:
+            env_file_dir = os.path.dirname(os.path.abspath(self.spack_env_file))
+            env_copy_cmd = "cp -r {0} {1}".format(env_file_dir, self.spack_env_directory)
+            res = sexe(env_copy_cmd, echo=True)
         if res != 0:
             print("[ERROR: Failed to create Spack Environment]")
             sys.exit(-1)


uberenv.py
```diff
@@ -902,11 +902,14 @@ def patch(self):
     def create_spack_env(self):
         # Create Spack Environment
         print("[creating spack env]")
+        res = 0
```
chapman39, line 905:
```suggestion
```

chapman39:
This doesn't appear to cause any issues in Serac, but I am worried there may be some consequence down the road. Does Spack not support moving external files to the Spack Environment that are included in the `spack.yaml`? Maybe we can create an issue in Spack.

chapman39:
@adrienbernede personally im ok with merging but @white238 @cyrush thoughts?

chapman39 (approved):

white238:
@becker33 thoughts?

becker33:
If you aren't using `--keep-relative` in your `spack env create` command, it's supposed to be de-relativizing the paths -- sounds like that isn't working.

I'm able to reproduce it not working locally, but I'm not sure that I'll have time to debug it this week.

becker33:
Can you try applying this diff to Spack and see whether the change is still necessary?

```
$ git diff
diff --git a/lib/spack/spack/environment/environment.py b/lib/spack/spack/environment/environment.py
index 7877a84764..37ef654888 100644
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -3015,7 +3015,7 @@ def included_config_scopes(self) -> List[spack.config.ConfigScope]:
         if missing:
             msg = "Detected {0} missing include path(s):".format(len(missing))
             msg += "\n   {0}".format("\n   ".join(missing))
-            raise spack.config.ConfigFileError(msg)
+            tty.warn(msg)
 
         return scopes
```

I have it still emiting a warning, a complete feature would avoid that but require a bit more engineering -- this should be enough to test whether it works for you and is worth moving forward.

becker33:
@adrienbernede when you use it with that change, what paths do you see in the created environment?

The reason I'm asking is that Spack is supposed to be turning them into absolute paths, which would get rid of the need to copy them at all. But it doesn't appear to be happening properly in your case, and I'm trying to figure out why.

adrienbernede:
@becker33 here is a reproducer with no uberenv involved:
``` console
## Radiuss Spack Configs
> git clone git@github.com:LLNL/radiuss-spack-configs.git
> git checkout woptim/config-updates
## Spack
> cd <path/to/spack>
> git checkout develop-2025-01-12
> . share/spack/setup-env.sh
## Let's go!
> spack env create -d ./spack_env <path/to/radiuss-spack-configs>/toss_4_x86_64_ib/spack.yaml
> cat <path/to/radiuss-spack-configs>/toss_4_x86_64_ib/spack.yaml
spack:
  include:
    - config.yaml
    - compilers.yaml
    - packages.yaml
  concretizer:
    reuse: false
    unify: false
  view: false
> cat spack_env/spack.yaml
spack:
  include:
    - config.yaml
    - compilers.yaml
    - packages.yaml
  concretizer:
    reuse: false
    unify: false
  view: false
```
I also tried using `./config.yaml`, etc, with no success.

becker33:
@adrienbernede can you try Spack with this patch and confirm it works for your use case? I intend to propose it to the rest of the Spack team once you confirm it works for you:

```
diff --git a/lib/spack/spack/environment/environment.py b/lib/spack/spack/environment/environment.py
index 7877a84764..d13f1edf3a 100644
--- a/lib/spack/spack/environment/environment.py
+++ b/lib/spack/spack/environment/environment.py
@@ -2634,6 +2634,29 @@ def _ensure_env_dir():
 
     shutil.copy(envfile, target_manifest)
 
+    # Copy relative path includes that live inside the environment dir
+    try:
+        manifest = EnvironmentManifestFile(environment_dir)
+    except Exception as e:
+        msg = f"cannot initialize environment, '{environment_dir}' from manifest"
+        raise SpackEnvironmentError(msg) from e
+    else:
+        includes = manifest[TOP_LEVEL_KEY].get("include", [])
+        for include in includes:
+            if os.path.isabs(include):
+                continue
+
+            abspath = pathlib.Path(os.path.normpath(environment_dir / include))
+            if not abspath.is_relative_to(environment_dir):
+                # Warn that we are not copying relative path
+                msg = "Spack will not copy relative include path from outside environment"
+                msg += f" directory: {include}"
+                tty.warn(msg)
+                continue
+
+            orig_abspath = os.path.normpath(envfile.parent / include)
+            shutil.copy(orig_abspath, abspath)
+
 
 class EnvironmentManifestFile(collections.abc.Mapping):
     """Manages the in-memory representation of a manifest file, and its synchronization
```

adrienbernede:
@becker33 I confirm that tests passed with this change and causes the present PR to become obsolete.

adrienbernede:
@becker33 Please let us know when the fix is pushed. I'll be happy to use the following snapshot release.

becker33:
Fix is here: https://github.com/spack/spack/pull/48689

chapman39:
Should this PR be closed?

adrienbernede:
@chapman39 I need for these branch to exist a little longer.
Updating Spack to the latest release with the fix led to bugs.
You may close it, but please do not remove the branch.

adrienbernede:
I don't need the PR anymore, and pushed a tag for reference of the commit if necessary.

PR #139, opened 2025-02-04
Revert `spack_setup_clingo` option back to using boolean values

chapman39:
I had mistakenly assumed that JSON `false` did not convert to Python `False` when decoding JSON files. This PR reverts the option back to using boolean type. Thank you @adrienbernede for pointing this out.


white238 (approved):

adrienbernede (commented):
Works for me. Thanks!

adrienbernede (approved):

PR #141, opened 2025-05-08
Adds logic to copy `versions.yaml` and `defaults.yaml` to spack environment

kennyweiss:
I'm using these files to consolidate shared versions (`versions.yaml`) and variants (`defaults.yaml`) for Axom's spack dependencies -- see https://github.com/LLNL/axom/pull/1574
In particular:
* `versions.yaml`: https://github.com/LLNL/axom/pull/1574/files#diff-45e557c81c8209d8dbeb4f52a0239eccc53c22a6c51ced39435349181ab3cbf7
* `defaults.yaml`: https://github.com/LLNL/axom/pull/1574/files#diff-5a7a33e81cb3fe713d044826e14c10cc58301354f60d2b5526e1e2068664a050
* and a platform env file: https://github.com/LLNL/axom/pull/1574/files#diff-88c7deb79649cce5aa1507c32ec80b76fcc3cbc0907283356600f1389dddd37e

This initial implementation might be too specific to Axom's use-case.
If so, we might want to add a new Uberenv config variable with a list of files top copy over, rather than hardcoding them.

I'm open to suggestions...

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -756,7 +756,10 @@ class SpackEnv(UberEnv):
           print("[checking for '{0}' yaml file]".format(_src))
           if os.path.exists(_src):
             print("[copying '{0}' config file to {1}]".format(_config_file, _dst))
-            sexe("cp {0} {1}".format(_src, _dst))
+            if os.path.exists(_dst):
+                raise FileExistsError(f"The file '{_dst}' already exists.")
+            else:
+                shutil.copy(_src, _dst)
 
         # If you still could not find a spack.yaml, create one later on
         if self.spack_env_file is None:
--- a/uberenv.py
+++ b/uberenv.py
@@ -544,7 +544,7 @@ class VcpkgEnv(UberEnv):
         dest_vcpkg_ports = pjoin(self.dest_vcpkg, "ports")
 
         print("[info: copying from {0} to {1}]".format(self.vcpkg_ports_path, dest_vcpkg_ports))
-        shutil.copytree(self.vcpkg_ports_path, dest_vcpkg_ports)
+        shutil.copytree(self.vcpkg_ports_path, dest_vcpkg_ports, dirs_exist_ok=True)
 
 
     def clean_build(self):
--- a/uberenv.py
+++ b/uberenv.py
@@ -756,10 +756,7 @@ class SpackEnv(UberEnv):
           print("[checking for '{0}' yaml file]".format(_src))
           if os.path.exists(_src):
             print("[copying '{0}' config file to {1}]".format(_config_file, _dst))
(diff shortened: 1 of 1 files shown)


cyrush:
This will be very helpful, thanks for working through a solution that allows you to reuse the common chunks of the configs.

chapman39 (approved):

kennyweiss:
> im testing this in serac now and don't expect to have too many issues. i think its going to clean things up a lot.
> 
> i was curious what the motivation was for separating `versions` and `defaults`. Why not combine them?

I agree that it's somewhat arbitrary and subjective. 

It was mostly to separate concerns. Over time, we're likely to change the versions more frequently than we are the default variants and this makes it easier to see when the versions changed.

I'd imagine that in the future, we'll either want to pass in the files to copy as a list (e.g. via the uberenv config file), or, ideally, spack will take care of the copies for us.

uberenv.py
```diff
@@ -748,6 +749,15 @@ def setup_paths_and_dirs(self):
...
                     self.spack_env_file = None
 
+        # Copy "defaults.yaml" and "versions.yaml" from configs dir, if they exist
+        for _config_file in ("defaults.yaml", "versions.yaml"):
+          _src = pjoin(spack_configs_path, _config_file)
+          _dst = pabs(pjoin(self.spack_env_directory, "..", _config_file))
+          print("[checking for '{0}' yaml file]".format(_src))
+          if os.path.exists(_src):
+            print("[copying '{0}' config file to {1}]".format(_config_file, _dst))
+            sexe("cp {0} {1}".format(_src, _dst))
```
white238, line 759:
```suggestion
            shutil.copy(_src, _dst)
```

kennyweiss: If we run uberenv a few times, will it always wipe out previous copy of `versions.yaml` and `defaults.yaml`?
I'll merge this and try it out, and revert if it's a problem.

white238: oh thats a good point... just change it to `shutil.copy`

kennyweiss: I confirmed that it threw the error on subsequent runs and removed the check:
```
[copying 'defaults.yaml' config file to .../uberenv_libs/defaults.yaml]
Traceback (most recent call last):
  File "..././scripts/uberenv/uberenv.py", line 1370, in <module>
    sys.exit(main())
  File "..././scripts/uberenv/uberenv.py", line 1305, in main
    env.setup_paths_and_dirs()
  File "..././scripts/uberenv/uberenv.py", line 760, in setup_paths_and_dirs
    raise FileExistsError(f"The file '{_dst}' already exists.")
FileExistsError: The file '.../uberenv_libs/defaults.yaml' already exists.
```

PR #142, opened 2025-07-07
add bootstrap defense

cyrush:
spack configs created with post 1.0 spack will undermine bootstrap of older spack (0.23.0) - here is an example of the error message you get:

```
[exe: /Users/harrison37/Work/github/llnl/conduit/_jhhh/spack/bin/spack -D /Users/harrison37/Work/github/llnl/conduit/_jhhh/spack_env concretize --fresh ]
==> Error: a single spec was requested, but parsed more than one:
gcc@14.2.0 languages:=fortran
          ^^^^^^^^^^^^^^^^^^^
```

Note the `:=` syntax is newer then spack 0.23.0

This adds `-b` to the our clean command, to re-bootstrap and avoid config issues from undermining builds.

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -980,7 +980,7 @@ class SpackEnv(UberEnv):
     def clean_build(self):
         # clean out any spack cached stuff (except build stages, downloads)
         # note: we now clean out spack bootstrap b/c ~/.spack contents can undermine bootstrap
-        cln_cmd = "{0} clean -b --misc-cache --failures --python-cache".format(self.spack_exe(use_spack_env=False))
+        cln_cmd = "{0} clean --bootstrap --misc-cache --failures --python-cache".format(self.spack_exe(use_spack_env=False))
         res = sexe(cln_cmd, echo=True)
 
         # check if we need to force uninstall of selected packages


uberenv.py
```diff
@@ -978,9 +978,9 @@ def concretize_spack_env(self):
         sexe(spack_concretize_cmd, echo=True)
 
     def clean_build(self):
-        # clean out any spack cached stuff (except build stages, downloads, &
-        # spack's bootstrapping software)
-        cln_cmd = "{0} clean --misc-cache --failures --python-cache".format(self.spack_exe(use_spack_env=False))
+        # clean out any spack cached stuff (except build stages, downloads)
+        # note: we now clean out spack bootstrap b/c ~/.spack contents can undermine bootstrap
+        cln_cmd = "{0} clean -b --misc-cache --failures --python-cache".format(self.spack_exe(use_spack_env=False))
```
white238, line 983:
```suggestion
        cln_cmd = "{0} clean --bootstrap --misc-cache --failures --python-cache".format(self.spack_exe(use_spack_env=False))
```
nitpick for consistency and clarity

white238:
We might hit a parallelism problem cleaning out the bootstrapping... I think the changing the bootstrap location is probably better.

cyrush:
bootstrap dir isn't enough, spack user cache dir needs to be controlled via SPACK_USER_CACHE_PATH, or hacked out.

In my test branch, I am updating the script to hard code the user cache path to be in a `spack_cache` dir under dest dir.
I think this actually may also solve the bootstrap issue, but I am not confident -- so I am keeping the bootstrap path set command in.

PR #149, opened 2025-08-26
Add spack package URL, tag, and branch overrides

white238:
I tested URL, tag, and branch but I get the following error with apple git:

```
[exe: /Users/white238/projects/uberenv/repo/uberenv_libs/spack/bin/spack repo update --tag develop-2025-07-13 builtin]
remote: Enumerating objects: 19569, done.
remote: Counting objects: 100% (19569/19569), done.
remote: Compressing objects: 100% (10512/10512), done.
remote: Total 19569 (delta 1300), reused 13548 (delta 1221), pack-reused 0 (from 0)
error: pathspec 'develop-2025-07-13' did not match any file(s) known to git
==> Error: Failed to update repository builtin
[ERROR: Failed to update git tag for builtin package repository]
```

I say we can merge this.

### DIFF
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -224,8 +224,11 @@ Project settings are as follows:
   spack_activate           **None**                   Spack packages to activate                       **None**
   spack_build_mode         ``--spack-build-mode``     Set mode used to build TPLs with Spack           ``dev-build``
   spack_configs_path       **None**                   Directory with Spack configs to be autodetected  ``spack_configs``
+  spack_packages_url       **None**                   Download url for Spack packages                  ``https://github.com/spack/spack-packages.git``
   spack_packages_path      **None**                   Directory|List with Package Repos to be added    ``packages``
+  spack_packages_branch    **None**                   Spack packages repo branch to checkout           **None**
   spack_packages_commit    **None**                   Spack packages repo commit to checkout           **None**
+  spack_packages_tag       **None**                   Spack packages repo tag to checkout.             **None**
   spack_setup_clingo       **None**                   Do not install clingo if set to ``false``        **None**
   spack_externals          ``--spack-externals``      Space delimited string of packages for Spack to  **None**
                                                       search for externals
--- a/docs/sphinx/index.rst
+++ b/docs/sphinx/index.rst
@@ -241,6 +241,8 @@ Project settings are as follows:
  ========================= ========================== ================================================ =======================================
 
 If a ``spack_commit`` is present, it supercedes the ``spack_branch`` option, and similarly for ``vcpkg_commit``and ``vcpkg_branch``.
+The precedence for the Spack packages repo options are ``spack_packages_commit``, ``spack_packages_branch``, and last ``spack_packages_tag``.
+If none of these are given, Spack will pull the current commit on their default branch.
 
 When used as a submodule ``.uberenv_config.json`` should define both ``spack_configs_path`` and ``spack_packages_path``,
 providing Uberenv with the respective location of ``spack_configs`` and ``packages`` directories.
--- a/uberenv.py
+++ b/uberenv.py
@@ -871,12 +871,18 @@ class SpackEnv(UberEnv):
 
             # Optionally, check out Spack's builtin package repo to a specific commit/branch/tag
             if "spack_packages_url" in self.project_args:
-                url = self.project_args["spack_packages_url"]
-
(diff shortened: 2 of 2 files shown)


docs/sphinx/index.rst
```diff
@@ -224,8 +224,11 @@ Project settings are as follows:
   spack_activate           **None**                   Spack packages to activate                       **None**
   spack_build_mode         ``--spack-build-mode``     Set mode used to build TPLs with Spack           ``dev-build``
   spack_configs_path       **None**                   Directory with Spack configs to be autodetected  ``spack_configs``
+  spack_packages_url       **None**                   Download url for Spack packages                  ``https://github.com/spack/spack-packages.git``
   spack_packages_path      **None**                   Directory|List with Package Repos to be added    ``packages``
+  spack_packages_branch    **None**                   Spack packages repo branch to checkout           **None**
   spack_packages_commit    **None**                   Spack packages repo commit to checkout           **None**
+  spack_packages_tag       **None**                   Spack packages repo tag to checkout.             **None**
```
chapman39, line 231: may be worth specifying the relationship between the spack_packages_x options somewhere. the fact you should only specify one of these, and/or which takes precedence.

chapman39 (approved):
ty chris

white238:
> it might be because when uberenv clones spack it's shallow or something?

Spack is controlling how it clones the spack-packages repo. Alec was going to look into it but it only happens on tags and in apple git as far as i can tell.

docs/sphinx/index.rst
white238: Done!

adrienbernede:
@white238 @chapman39 I am interested in this feature, but there is a native way to specify the spack-packages commit/branch/tag in a spack environment.
When I try that, the commit specified appears to be ignored.
- I am looking into why this is ignored. If you know, get in touch.
- How is uberenv supposed to cope with 2 source of truth ? I guess the .uberenv-config.json will have precedence. We should state it somewhere.
My ideal scenario would be that both are supported, and I could choose to only rely on the environment file if I’d like.

chapman39:
@adrienbernede how do you specify it via spack environment? if there's a way to specify it in the spack.yaml/ spack environment, its possible we don't need to via uberenv.

white238:
My experience was there were multiple commands that force downloading of the package repo. `spack bootstrap now` was one of them. We should find a way to enable both options.

alecbcs:
@chapman39 to specify a specific branch, tag, or commit for a package repository you can add the following to your Spack environment. We had a bug in v1.0 that would always download `spack/spack-package@develop` on the very first run of Spack even when another tag was specified. We fixed that in v1.0.1.

```yaml
spack:
  repos:
    builtin: # repository name
      git: https://github.com/spack/spack-packages.git
      branch: develop
      # tag: v2025.07.0
      # commit: abc123
      destination: /path/to/download/location
```

adrienbernede:
@alecbcs in Uberenv I had to run `spack repo update`, here is why:
- We first run spack before setting the environment to so the commit we get for the spack-packages in not the one we want.
- Then we run `spack env create` to copy the spack config file at the desired location. This does not update the spack-packages ref.
- Then we don't actually activate the environment, but instead run `spack -D /path/to/env <command>`. And this does not update the spack-packages ref neither.

This is why, after the env creation, I had to run `spack repo update`. I think that's fine, but I thought you should know.

adrienbernede:
> It is fine if we support both options but I personally don't want the spack package information in multiple places. It makes the most sense in my opinion to be in the uberenv config file

While I understand your reasons, I think this is not going in the direction of maintaining stand alone environments.
I would like my environments to be like any spack environment: if the environment is tied to a spack-packages ref, this ref should be in the spack.yaml file.

That's why I vote for spack-packages ref in the spack.yaml file only.

If you are going to support refs set in the uberenv config, my proposal would be that the ref can be set in both, but Uberenv would warn if both are different. That's more complex, potentially breaking, potentially confusing. I don't like it...

What is settable in an environnement should be in the environment, that’s my take.

white238:
The main use case of this, and how it worked previously, was to have uberenv setup Spack in a singular blessed git state at the project level.  Having it defined in the environment file works as long as you have existing environments but for a lot of my users, they don't have an environment file yet because they are doing development on their local machine. 

Again I am not against having these as options, it currently is optional. But it doesn't work for my projects to require it in the environment file.

PR #154, opened 2025-12-17
Prevent mixing compilers in Spack versions over 1.1.0

chapman39:
Adds option `--spack-allow-compiler-mixing`. If left unset, Uberenv will disable compiler mixing by default.

More info:

* https://spack.readthedocs.io/en/latest/configuring_compilers.html#mixing-compilers
* https://github.com/spack/spack/pull/51135

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -63,7 +63,6 @@ import glob
 import re
 import argparse
 
-from packaging.version import Version
 from functools import partial
 
 from os import environ as env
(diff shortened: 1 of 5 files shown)


white238:
> Fix in [spack/spack#51796](https://github.com/spack/spack/pull/51796)

This has been merged. We should move this forward now.

uberenv.py
```diff
@@ -985,6 +1009,17 @@ def create_spack_env(self):
             print("[ERROR: Failed to update git reference for builtin package repository]")
             sys.exit(-1)
 
+        if version2tuple(self.spack_version()) >= version2tuple("1.1.0"):
```
white238, line 1071: Instead of `version2tuple`, I would recommend just using this:

```
from packaging.version import Version

...

if Version(self.spack_version()) > Version("1.1.0"):
```

chapman39: im remembering now why i didn't use that: apparently it doesn't work on mac without `pip install packaging`, and i believe @cyrush wants to avoid adding dependencies to uberenv.

https://github.com/llnl/uberenv/actions/runs/29538891573/job/87756581427?pr=154#step:3:7

chapman39:
i tested this again on lido and seems to be working. once i switch out the Version usage should be ready for review again.

uberenv.py
```diff
@@ -313,6 +313,13 @@ def parse_args():
                       default=None,
                       help="Path to Spack Environment file (e.g. spack.yaml or spack.lock)")
 
+    # Enable compiler mixing argument (Spack 1.1.0+ only)
+    parser.add_argument("--spack-compiler-mixing",
+                      dest="spack_compiler_mixing",
+                      default=False,
```
white238, line 319: Personally I think this should default to true

white238 (approved):

uberenv.py
chapman39: if thats the case then i changed the argument to `--spack-disable-compiler-mixing`

that way if left unset, uberenv won't override spack's default settings or what's already in a users's spack environment.

white238: We chatted directly and are going with `--spack-allow-compiler-mixing` which defaults to `false`, so compiler mixing is off by default.

white238 (approved):

PR #160, opened 2026-06-19
Enables setting vcpkg package features

kennyweiss:
* This branch adds the ability to enable vcpkg features through the command line via `--vcpkg-features` (roughly equivalent to spack variants).
  * The vcpkg syntax is `vcpkg install foo[bar]` to enable feature `bar` for package `foo`
  * I modified the install line to include `--recurse` which allows reinstalling already built dependencies
  * We also now check the return code from installation and print and error message if it is nonzero.
* I also added a command line argument to set the CUDA arch flags: `--vcpkg-cuda-architectures`.

Given the above two features, I am using this generate vcpkg TPLs for a cuda-enabled axom (in a forthcoming PR to Axom).

### DIFF
--- a/uberenv.py
+++ b/uberenv.py
@@ -155,6 +155,18 @@ def parse_args():
                       default=None,
                       help="dir with vckpkg ports")
 
+    # optional vcpkg feature list for the package
+    parser.add_argument("--vcpkg-features",
+                      dest="vcpkg_features",
+                      default=None,
+                      help="comma-separated vcpkg features to enable for the package")
+
+    # optional CUDA architecture list for vcpkg CUDA builds
+    parser.add_argument("--vcpkg-cuda-architectures",
+                      dest="vcpkg_cuda_architectures",
+                      default=None,
+                      help="semicolon- or comma-separated CUDA architectures for vcpkg CUDA builds")
+
     # overrides package_name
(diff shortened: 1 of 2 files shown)


chapman39 (approved):
might be good to list these new options in the docs

uberenv.py
```diff
@@ -155,6 +155,18 @@ def parse_args():
...
+    parser.add_argument("--vcpkg-features",
+                      dest="vcpkg_features",
+                      default=None,
+                      help="comma-separated vcpkg features to enable for the package")
+
+    # optional CUDA architecture list for vcpkg CUDA builds
+    parser.add_argument("--vcpkg-cuda-architectures",
+                      dest="vcpkg_cuda_architectures",
+                      default=None,
+                      help="semicolon- or comma-separated CUDA architectures for vcpkg CUDA builds")
```
white238, line 168: nitpick: is there a benefit to supporting both comma and semi-colon delimited list? Even if vcpkg supports both maybe we just list one here?

white238 (approved):

uberenv.py
kennyweiss: That's fair -- I'll figure out which one makes more sense and update it

